PR #4, opened 2017-06-02
Allow open-service-broker to accept coma-separated ip ranges in its config

Lafunamor:
This change will allow the open-service-broker to parse the IP field and return multiple security groups for each range.


mcelep:
There is a way to parse array of objects from yml directly into list of objects into the code.
Instead of doing the parsing yourself, it would perhaps make sense to use existing mechanisms and add a comment on the configuration.

Lafunamor:
close PR in favor of https://github.com/swisscom/open-service-broker/pull/6

PR #8, opened 2017-06-21
Changed DELETE method to look for path variable 'service_id' instead of 'id'

Zee-Khan:


ampersand8:
Hey Zee, can you please add the test for this use case?

meip:
We added a test and fixed the issue. Thank you for bringing it up!

PR #25, opened 2017-08-07
Feature/add semi mutual auth certs in yml

konradlykowski:


broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/kubernetes/config/KubernetesConfig.groovy
```diff
@@ -15,6 +15,12 @@ import org.springframework.context.annotation.Configuration
 class KubernetesConfig implements Config, EndpointConfig, AsyncServiceConfig {
     String kubernetesHost
     String kubernetesPort
-    String kubernetesClientPFXPath
-    String kubernetesClientPFXPasswordPath
+    String kubernetesClientCertificate
+    String kubernetesClientKey
+    String kubernetesClientKeyStorePassword
```
MatthiasWinzeler: why is `kubernetesClientKeyStorePassword`, `kubernetesClientCertAlias` and `kubernetesClientKeyAlias ` needed to be configured? AFAIK it's an internal thing that the operator does not have to worry about...

broker/src/main/resources/application.yml
```diff
@@ -66,8 +66,6 @@ com.swisscom.cloud.sb.broker:
     kubernetes:
       kubernetesHost:
       kubernetesPort:
```
MatthiasWinzeler: shouldn't the new config items be exposed in application.yml?

PR #26, opened 2017-08-15
Feature/cloudac 6458 service definition management

Zee-Khan:


broker/build.gradle
```diff
@@ -10,6 +10,23 @@ buildscript {
...
+        classpath "gradle.plugin.com.github.jk1:gradle-license-report:0.3.12"
+    }
+}
+
+apply plugin: "com.github.jk1.dependency-license-report"
+
+licenseReport {
+    renderer = new com.github.jk1.license.render.TextReportRenderer("THIRD-PARTY-NOTICES.txt")
+}
+
```
mcelep: Does not  the style below of specifying the plugin work?:
`
plugins {
  id "com.github.jk1.dependency-license-report" version "0.3.12"
}
`

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.ApplicationArguments
+import org.springframework.boot.ApplicationRunner
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.stereotype.Component
+
+@Component
+@EnableConfigurationProperties
+@Slf4j
+class ServiceDefinitionInitializer implements ApplicationRunner{
```
mcelep: Why do we opt for this interface instead of declaring a method that is annotated with `@PostConstrcut` ?

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    void checkForMissingServiceDefinitions(List<CFService> cfServiceList) {
+        def configGuidList = serviceDefinitionConfig.serviceDefinitions.collect {it.guid}
+
+        def guidList = cfServiceList.collect {it.guid}
+
+        if (configGuidList.size() != 0) {
+            if (!configGuidList.containsAll(guidList)) {
+                throw new Exception("Missing service definition configuration exception. Service list - ${guidList}")
```
mcelep: Should use a runtime Exception?

README.md
```diff
@@ -189,7 +183,40 @@ Check the class [`BoshBasedServiceProvider`](broker/src/main/groovy/com/swisscom
...
+Any [Kubernetes](http://kubernetes.io) based service can be provisioned with Open Service Broker. The asynchronous task is being created to prepare the provisioning of the service instance. Kubernetes [Facade](https://github.com/swisscom/open-service-broker/blob/develop/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/kubernetes/facade/KubernetesFacade.groovy) is using the [client](https://github.com/swisscom/open-service-broker/blob/develop/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/kubernetes/client/rest/KubernetesClient.groovy) to execute a bunch of "templated" HTTP calls on Kubernetes Server. All the templates are automatically read from provided directory and matched with k8s endpoint.
      
+### OpenWhisk
+Open Service Broker can broker your local OpenWhisk deployment.
```
Zee-Khan: Local is needed as our brokerage of OpenWhisk won't work with their production version as we don't have access to the database. I guess local could be misleading as it doesn't need to be on your laptop. Maybe I should say "your own deployment of OpenWhisk"

README.md
```diff
@@ -189,7 +183,40 @@ Check the class [`BoshBasedServiceProvider`](broker/src/main/groovy/com/swisscom
...
+
+Provision will create a new namespace.
 
+An example of the provision json data (Parameters are optional).
```
Zee-Khan: Although I work on the Service Broker, I don't think I ever used the cf cli to provision/bind a service. Will update.

broker/build.gradle
Zee-Khan: I think it does. The way I wrote it was compatible with all versions so I thought it was safer. I can use the new notation.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
Zee-Khan: When looking into the best way to perform an action after the application was running, this was the suggested interface. For @PostConstruct, would I create a method in the ServiceBroker class?

README.md
Zee-Khan: Now that I think of it, I probably could have brokered against the production OpenWhisk. The logic would different and probably not need access to the database. We'd just need to differentiate between local/public OpenWhisk.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
mcelep: No, you would create a method called for ex. 'init' in ServiceDefinitionInitializer  and annotate it with PostConstruct.

README.md
mcelep: I would love to understand more more about: `The logic would different and probably not need access to the database.`

README.md
mcelep: let's keep both then. Let's have your json and also a cf cli example to make it clear about what parts are optionals.

broker/build.gradle
mcelep: We focus gradle/wrapper version that we have and don't worry for version older than 2.1 (which is when the plugins mechanism was introduced based on t[his](https://plugins.gradle.org/plugin/com.github.jk1.dependency-license-report))

README.md
Zee-Khan: When asking the OW team about CRUD on namespaces/subject, they said that the open sourced version doesn't have it built into the API so we directly updated the DB to work around this. Maybe their public Bluemix offering does have the API enabled for CRUD on namespaces/subjects. If this is the case, then we'd provision/bind differently as we'd be able to do it through the API. Doing a quick search, however, shows that even the public offering only allows reading of namespaces/subjects through the API.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
Zee-Khan: Done.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ServiceDefinitionInitializer.groovy
Zee-Khan: Done.

README.md
```diff
@@ -186,11 +186,16 @@ https://github.com/swisscom/mongodb-enterprise-boshrelease
...
-Open Service Broker can broker your local OpenWhisk deployment.
+Open Service Broker can broker your local deployment of OpenWhisk.
 
 Provision will create a new namespace.
 
 An example of the provision json data (Parameters are optional).
+```text
+cf create-service SERVICE PLAN SERVICE_INSTANCE -c '{"name":"value","name":"value"}'
+ 
+cf create-service udn9276f-hod4-5432-vw34-6c33d7359c20 jfkos87r-truz-4567-liop-dfrwscvbnm20 59ee663f-a964-4a28-8315-47634300ad37 -c '{"namespace":"NAMESPACE"}'
```
mcelep: those three arguments you pass to cf create-service are usually more meaningful e.g.:
cf create-service openwhisk usage my-open-whisk
cf create-service openwhisk usage my-open-whisk -c '{"namespace":"NAMESPACE"}'

PR #27, opened 2017-08-16
Feature/issue19/refactor internal name

### LINKED ISSUE #19: Refactor field internalName in CFService
The field `internalName` in `CFService` and `Plan` is used to reference a corresponding `ServiceProvider`. This name is somehow misleading when a ServiceDefinition got pushed (as seen in #9).
I suggest to rename these fields to `serviceProviderName`. This migration isn't trivial since backwards compatibility should be given. I suggest the following migration path:

1. Introduce a new field with the appropriate name
1. Add a mapping functionality to map the old field name to the new one
1. Remove this mapping functionality with the next major release

Zee-Khan:
Fixes #19. May have conflicts with service definition branch.


broker/src/main/groovy/com/swisscom/cloud/sb/broker/model/CFService.groovy
```diff
@@ -13,6 +13,7 @@ class CFService extends BaseModel{
     String description
     Boolean bindable
     String internalName
+    String serviceProviderClass
```
mcelep: Please make sure that classes such as ServiceLifeCycler which reference this field are also updated

PR #30, opened 2017-08-25
use separate cert and key files

ampersand8:
From an operational point of view it's easier to manage certificates and keys in separate files.


meip:
I can also see advantages when everything can be configured from the application.yml:
+ only 1 source for errors vs. file systems errors (file doesn't exists, permissions, file has wrong content etc.)
+ independent from the deployment platform

ampersand8:
I can see the benefits of having just one file to deal with, but in my experience those are largely visible during development. For operations there are many advantages to put the certs and keys in separate files:

- During production deployment there will need to be some sort of application.yml file, so the filesystem problems have to be dealt with anyway.
- All common plattforms allow to have config files around (path should be configurable though).
- For operators a separate cert file is very easy to handle (replace, copy, validate with openssl).
- Having certs and keys in separate files seems to be convention for bosh-releases, as most of the projects handle it this way.
- Removing certs and keys from the config file makes it a bit more readable, as otherwise one has to skip around to see different values. Generally speeking, having a multi thousand line config file is not convenient to quickly change something.

ampersand8:
Thanks @MatthiasWinzeler, overall I have to agree with you.

PR #32, opened 2017-09-11
Feature/k8s redis shield registration

meip:
*Description*
As a ServiceBroker for Kubernetes based service I want to configure a "system-driven" backup for each service provisioning in a 6 hours interval so that the latest backup can be restored in case of a failure or disaster.

*Details*
When a new service on Kubernetes got provisioned _OSB_ has to configure new backup for every 6 hours. Once the backup is configured _OSB_ relays on the backup service to have consistent backups.
When a service got deprovisioned also the backup configuration has to be removed and all backups of this service should get destroyed immediately. _OSB_ doesn't have to keep track of any records for system driven backups after a service got deprovisioned. The configuration parameters for the backup can be set from the service definition as followed `BACKUP_SCHEDULE_NAME`, `BACKUP_POLICY_NAME`, `BACKUP_STORAGE_NAME`.


MatthiasWinzeler:
Good job! Just some small remarks from my side:

- It seems when migrating the Shield part from the old SB, the newest changes were not copied to the new codebase (there were some commits made to the old SB in the meantime). Can you add these? Just check the latest commits on these files in the old repo.
- I think we should rename `createShieldTarget`; it sounds like it's creating stuff (i.e. on a 3rd party system) whereas it's just building an object. What about `buildShieldTarget`?
- What prevents you to merge `ShieldClient` and `ShieldClient_2_3_5` into one class?
- What happens if the required params in the service def are not provided, does it fail on deployment time (I think that would be nice)?

meip:
Thank you @MatthiasWinzeler for the feedback. 
* I synced again the Shield parts with the old repo.
* I like the renaming suggestion as well and renamed it to `buildShieldTarget`
* Since the change to get `BACKUP_SCHEDULE_NAME`, `BACKUP_POLICY_NAME`, `BACKUP_STORAGE_NAME` from the _Service Definition_ a bigger refactoring on `ShieldClient` and `ShieldBackupRestoreProvider` would be required. I didn't found time go follow up on this and thus pushed the `ShieldClient_2_3_5` as an intermediate solution to don't break compatibility.
* Right now if one of the parameter is invalid (`null` or couldn't be found on Shield) a `RuntimeException` is thrown. Do you think this is sufficient?

MatthiasWinzeler:
@meip Thanks! As discussed, let's try to refactor `ShieldClient` now. For the startup, we'll verify this in a separate task.

PR #37, opened 2017-11-21
All service instance provision requests are saved

ampersand8:
**Motivation**
While testing kubernetes services, we experienced issues with some API calls. These led to deployment failures, which could not be resolved by requesting the service instance deletion, as OSB had not saved the service instance in the database. For async services this is done after the first step, which already partly failed.

**Description**
With this change, the service instance gets saved whether it will succeed or not.

### DIFF
--- a/broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/FirstAsyncProvisioningStepFailsFunctionalSpec.groovy
+++ b/broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/FirstAsyncProvisioningStepFailsFunctionalSpec.groovy
@@ -1,0 +1,31 @@
+package com.swisscom.cloud.sb.broker.functional
+
+import com.swisscom.cloud.sb.broker.services.common.ServiceProviderLookup
+import com.swisscom.cloud.sb.broker.util.test.DummyFailingServiceProvider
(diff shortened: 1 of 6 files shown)


MatthiasWinzeler:
Looks safe to me.
@meip What do you think?

@ampersand8 Have you verified this fixes the issue we had? Does deprovisioning k8s services now work if it fails during early stages of provisioning? 
Maybe you could even create an integration test for that...

ampersand8:
Yes, manual testing showed that the issue is solved for k8s deprovisioning.

meip:
I can follow your bug description and I would nail it down to this line
https://github.com/swisscom/open-service-broker/blob/1674e8d81e48b99c56904c723e1a37641d9bfb94/broker/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/job/ServiceProvisioningJob.groovy#L42
Here the provisioning job gets kicked off before it could have been saved which will happen on the next line
https://github.com/swisscom/open-service-broker/blob/1674e8d81e48b99c56904c723e1a37641d9bfb94/broker/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/job/ServiceProvisioningJob.groovy#L43

So as a fix I suggest to just switch this two lines (42 and 43) so that we can keep the whole async provisioning logic with the "job"-concept.

PR #38, opened 2017-11-21
log all RestTemplate calls

ampersand8:
**Motivation**
Individual REST calls to the Kubernetes API can fail for many reasons. Because we send at least six REST calls the probability that one of them is failing is a bit higher.

**Description**
With Spring Retry we can easily retry the individual failing REST calls.


MatthiasWinzeler:
I'm against introducing something unless we know it fixes a problem we know we have... otherwise, it might cause more harm than it helps (i.e. since `Retryable` will retry on *any* exception - not only exception we know we can safely retry because they're idempotent - it might cause weird issues.

meip:
For me this change is reasonable because some errors during API communication are worth a retry. I suggest to only retry on `HttpServerErrorException`. 

@MatthiasWinzeler why is `Retryable` retrying on any exception?

MatthiasWinzeler:
I assumed that `Retryable` retries on any exception by default because you can optionally pass an exception class so that it only retries on this class. 

`HttpServerErrorException` are all HTTP 5xx. Not all 5xx codes are semantically retryable (i.e. 502 and 503 are, 500 is not). Also, there are other cases that are retryable - i.e. Connection timeouts and resets.

I still think we should only implement this if it fixes an issue we have :)

ampersand8:
I agree with restraining from adding complexity for problems that we're not sure how relevant they're going to be. I would then suggest to add an identifiable log message for Kubernetes API responses (https://github.com/swisscom/open-service-broker/blob/1674e8d81e48b99c56904c723e1a37641d9bfb94/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/kubernetes/client/rest/KubernetesClient.groovy#L33), which we then can use in elasticsearch to gather information about the reliability.

MatthiasWinzeler:
@ampersand8 Good idea! Maybe we can just enable logging on **all** outgoing HTTP requests (more precisely: the status codes of their respective responses, even in success cases) - I guess this information is helpful in many cases. Do we have a way to enable this on all `restTemplates` in a generic, elegant way?

ampersand8:
@MatthiasWinzeler yes logging for RestTemplate calls can easily be added with an interceptor - see 
https://github.com/swisscom/open-service-broker/pull/38/commits/dc29b2f081c809f1f0f1c3c4bc124b01355ed01f

PR #47, opened 2018-01-29
Feature/support multiple application users

kristianszoke:


broker/src/main/resources/application.yml
```diff
@@ -22,11 +22,13 @@ info:
 
 server.tomcat.accesslog.enabled: true
 
-com.swisscom.cloud.sb.broker.security:
-  cfUsername: 'cc_admin'
-  cfPassword: 'change_me'
-  cfExtUsername: 'cf_ext'
-  cfExtPassword: 'change_me'
+com.swisscom.cloud.sb.broker.security.applicationUsers:
```
meip: Maybe we should introduce a `guid` which represents a platform so that it can be shared across many systems. This means a platform would be at least a `guid` with _n_ `CF_ADMIN` and _n_ `CF_EXT_ADMIN`. What do you think @kristianszoke and @MatthiasWinzeler ?

meip:
LGTM 👍 
@ampersand8 Please have a look as well shortly and then we'll merge it

ampersand8:
LGTM

PR #52, opened 2018-02-22
Feature/cloudification

seilc1:
* Added dependency to cloudfoundry to auto detect database in a cloud environment.
* Added manifest-template for cf
* Added gradle configuration to automatically build a manifest for cf to push with current build version


MatthiasWinzeler:
@seilc1 thanks, that's stuff we wanted for ages :)
quick question:
> Add CloudFoundry Connector, which automatically changes the datasource for the database connection in the cloud.
There's not only one DB connection (for SB's main db), but at least one more that I know of: The metrics collection db. Since this is a different one it should probably not be autoconfigured to be the main db.
@meip @ampersand8 can you help to verify it works with our multiple DBs?

seilc1:
I used https://spring.io/blog/2015/04/27/binding-to-data-services-with-spring-boot-in-cloud-foundry as a base. It already provides a 'solution' for the Problem. but it involves quite a bit of work to do so:

* Create special dataSources to be injected. with different profiles.
* Make the dataSources configurable from .yaml so that we can set one dataSource for the OSB database and one for the metric database (e.g. dataSource.OSB.ServiceName = 'broker'; dataSource.Metrics.ServiceName = 'metrics' )

MatthiasWinzeler:
@seilc1 Sounds good. Would you implement it as shown in your example (`dataSource.OSB.ServiceName = 'broker'`) with using the service name to distinguish the two service bindings? 

There might be also this [declarative approch](https://spring.io/blog/2015/04/27/binding-to-data-services-with-spring-boot-in-cloud-foundry#a-purely-declarative-approach) - could we use this so we don't have to write custom code to make the data sources configurable from application.yml?

Btw. was autoreconfiguration working for you on the Appcloud you were testing? There was a known issue because the service were missing tags, wonder if that's resolved.

meip:
@MatthiasWinzeler I think we're fine for the time being to only support 1 datasource configured through CloudFoundry Connector. The metrics collection DB doesn't make use of JPA functionality and is using direct JDBC setup.

MatthiasWinzeler:
@meip 👍

PR #53, opened 2018-02-26
Feature/fix application user initializer

seilc1:


broker/src/main/groovy/com/swisscom/cloud/sb/broker/security/ApplicationUserInitializer.groovy
```diff
@@ -17,20 +17,46 @@ import javax.annotation.PostConstruct
...
     @PostConstruct
     void init() throws Exception {
+        checkForDuplicatedApplicationUserConfigurations()
         synchronizeApplicationUsers()
     }
 
+    void checkForDuplicatedApplicationUserConfigurations()
+    {
+        def duplicatedUserConfigurations = new ArrayList<UserConfig>()
+        applicationUserConfig.platformUsers.each {
```
MatthiasWinzeler: cosmetic: wouldn't a `findAll` instead of an `each` be more elegant?

broker/src/main/groovy/com/swisscom/cloud/sb/broker/security/ApplicationUserInitializer.groovy
```diff
@@ -41,20 +67,17 @@ class ApplicationUserInitializer {
         def allDbUser = userRepository.findAll();
 
         applicationUserConfig.platformUsers.each {
```
MatthiasWinzeler: I guess we can replace every `each` with a more elegant method :)

broker/src/test/groovy/com/swisscom/cloud/sb/broker/servicedefinition/ApplicationUserInitializerSpec.groovy
```diff
@@ -0,0 +1,148 @@
+package com.swisscom.cloud.sb.broker.servicedefinition
```
MatthiasWinzeler: yippie, tests, thanks 👍

MatthiasWinzeler:
stupid question: Since the username is not unique anymore - I guess two platforms could have users with the same username. How do we then decide which platform it belongs to when the platform only sends the username in the basic auth creds?

broker/src/main/groovy/com/swisscom/cloud/sb/broker/security/ApplicationUserInitializer.groovy
meip: Having `each` here is okay since its causing a side-effect.

MatthiasWinzeler: makes sense.

seilc1:
There is no such thing as a stupid question. The Username is still unique. because of the problem you mentioned. It's a bit confusing as the change was reverted.

MatthiasWinzeler:
@seilc1 nice, LGTM!

kristianszoke:
nice, LGTM :)

PR #62, opened 2018-03-14
fix race condition in Kubernetes Redis provision

ampersand8:
Writing to injected objects is not thread safe.
This fix makes a copy of the injected map and pushes the remaining maps to the copy.


MatthiasWinzeler:
@ampersand8 nice catch! 
The pleasures of mutability... :) 

As an object, modifying an object that does not even belong to yourself (but to a reference of you) is quite a code smell and can lead to issues like this. Some languages go even so far to not enforce immutability of data structures like this, to avoid such issues.

Do we have any other places where we do this?
It should pretty easy to search for the hashmap operator `<<` and then check whether we modify a foreign object.

ampersand8:
While quickly skimming through the code I did not see any other obvious violations. But it might makes sense to add functional tests that do multiple requests simultaneously for all the services.

PR #63, opened 2018-03-14
added SQL script to create missing IDX and PK to service_instance_ser…

kristianszoke:
…vice_detail and service_binding_service_detail


GETandSELECT:
@kristianszoke I am just curious why here a procedure is needed. Why you can't just write `ALTER TABLE`? I don't see the added value of procedure. 

Thanks for feedback.

meip:
You're completely right that these simple `ALTER TABLE` commands could be executed without any wrappers around it. I don't have any preference here.

PR #64, opened 2018-03-21
Feature/extensions

Zee-Khan:


client/src/main/groovy/com/swisscom/cloud/sb/client/model/Extension.groovy
```diff
@@ -0,0 +1,13 @@
+package com.swisscom.cloud.sb.client.model
+
+import groovy.transform.CompileStatic
+
+@CompileStatic
+class Extension {
+    String discovery_url
+    String server_url
+    Map<String, Map<String, String>> credentials
+    String altInstanceId
```
ampersand8: What is altInstanceId referring to?

Zee-Khan: It was a field that was required in the previous version of the extensions proposal by the OSBA team that I thought I removed. It is removed now.

PR #67, opened 2018-03-23
Hotfix/backup retry

Zee-Khan:


broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -16,12 +18,15 @@ class ShieldClient {
     protected ShieldConfig shieldConfig
     protected ShieldRestClientFactory shieldRestClientFactory
     protected RestTemplateBuilder restTemplateBuilder
+    protected int retryBackupCount = 0
```
meip: Please refactor this member so that the bean stays stateless.

Zee-Khan: Done.

PR #76, opened 2018-04-09
openshift deployment script

mcelep:
A script to deploy osb on Openshift


meip:
Very nice, thx for the contribution! Would you like to add a small comment to the `README` as well?

PR #87, opened 2018-04-21
Renamed db migration file to avoid revision conflict

arcenik:
DB Migration is not wokring because two migrations file has the same revision id.

    -rw-r--r-- 1 fs fs  1165 21 avr 14:58 V1_0_18__add_applicationUser_owner_ServiceInstance.sql
   -rw-r--r-- 1 fs fs   568 21 avr 14:58 V1_0_18__add_retry_backup_count.sql


meip:
Good catch, thank you! 

I proposed to rename the file `V1_0_19__add_applicationUser_owner_ServiceInstance.sql` since this not yet part of a released version.

PR #89, opened 2018-04-26
Feature/general sb

melanielc:


mcelep:
Hey @melanielc, good to see that you're working on OSB. 
Some feedback:

- For comments,  how about going for imperative mood(instead of past tense)?. More details [here](https://chris.beams.io/posts/git-commit/#imperative).

In a nutshell:  

> A properly formed Git commit subject line should always be able to complete the following sentence:
> 
> If applied, this commit will your subject line here

- There are more than 2000 lines added in this branch, I understand the focus is on testing it would however make sense to get some feedback from others earlier. Trying to keep the batch size of commits & number of lines changed small, specially when new in a project, can save you time. If you got the feedback already and if it happens to be not visible to outside, then just ignore this comment :)

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceBrokerServiceProviderFunctionalSpec.groovy
```diff
@@ -0,0 +1,114 @@
+package com.swisscom.cloud.sb.broker.functional
+
+import com.swisscom.cloud.sb.broker.model.CFService
+import com.swisscom.cloud.sb.broker.model.Plan
+import com.swisscom.cloud.sb.broker.services.common.ServiceProviderLookup
+import com.swisscom.cloud.sb.broker.services.genericserviceprovider.TestableServiceBrokerServiceProvider
+import com.swisscom.cloud.sb.broker.util.test.DummyServiceProvider
+import com.swisscom.cloud.sb.client.model.LastOperationState
+
+class ServiceBrokerServiceProviderFunctionalSpec extends BaseFunctionalSpec {
```
meip: This functional test only works on an empty database because some static service instance guids are used, right? (see `TestableServiceBrokerServiceProvider`) Could be we may think of a solution for it?

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/genericserviceprovider/client/ServiceBrokerServiceProviderRestClient.groovy
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.stereotype.Component
+import org.springframework.web.client.RestTemplate
+
+@Component
+@CompileStatic
+@Slf4j
+class ServiceBrokerServiceProviderRestClient {
+
+    // maybe put url into sbspConfig?
+    private ServiceBrokerServiceProviderConfig sbspConfig
```
meip: Is this field still required?

PR #130, opened 2018-06-28
Support BOSH uaa authentication

ampersand8:
First call to BOSH API invokes a check for authentication method. If authentication method is `uaa` a token will be fetched and stored. Further calls will use the saved token unless it's expired.

### DIFF
--- a/broker/src/integration-test/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClientTest.groovy
+++ b/broker/src/integration-test/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClientTest.groovy
@@ -41,7 +41,7 @@ class BoshRestClientTest extends BaseSpecification {
 
     def "get deployment"() {
         when:
-        def deployment = client.getDeployment("d-8tlihzux5s3nq370")
+        def deployment = client.getDeployment("shield")
         then:
         noExceptionThrown()
     }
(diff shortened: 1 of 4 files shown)


MatthiasWinzeler:
@ampersand8 nice job! 

One little detail: I remember @mcelep once telling me that I should not store a token on a (potentially shared) object until it's expired but always get a new one. Otherwise, I might run into concurrency issues...

I wonder whether we should to it the same way here (always get a new token, no matter whether it's expired). I guess we don't really expect that much interaction with the BOSH director that we really need to optimize for reusing the token.

ampersand8:
That would simplify the code a bit. :+1:  Shouldn't this be thread safe, as it is not shared? I'm not sure, would need to write a test for that, but making `checkAuthTypeAndLogin` synchronized would probably help here.

We potentially have quite some calls to BOSH, looking at MongoDB Enterprise provisioning as an example:

2 Calls: Fetch and Update Cloud Config
1 Call: Create new deployment
32 Calls: Check deployment task state (if it takes 8 Minutes and we check by default every 15 seconds)

So total maybe **35 API Calls** (and therefore 35 tokens) plus everytime getting `/info` and `/oauth/token` as well. Could that be a problem?

With an optional authentication configuration we could get rid of the `/info` step, e.g.:
```
boshAuthentication:
  type: 'uaa'
  url: 'https://xyz'
```
But as you stated I'm not sure whether having one REST call less, is worth the added complexity for everyone involved.

MatthiasWinzeler:
@ampersand8 I think UAA can easily handle this kind of amount. It will also not be everyday that we provision VMs.

By the way, it's also done like this (never store the token) in the `UaaClient` (in osb-extended), you might find some inspiration there.

ampersand8:
@MatthiasWinzeler ok then it makes sense to always get a new token. I changed the code accordingly.

PR #137, opened 2018-07-25
Move RelationalDb functionality into a facade class

ampersand8:
For easier reuse of the RelationalDb the functionality is being moved to a facade class.

### DIFF
--- a/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/relationaldb/RelationalDbFacade.groovy
+++ b/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/relationaldb/RelationalDbFacade.groovy
@@ -1,0 +1,119 @@
+package com.swisscom.cloud.sb.broker.services.relationaldb
+
+import com.swisscom.cloud.sb.broker.binding.BindRequest
+import com.swisscom.cloud.sb.broker.binding.BindResponse
+import com.swisscom.cloud.sb.broker.binding.UnbindRequest
+import com.swisscom.cloud.sb.broker.error.ErrorCode
+import com.swisscom.cloud.sb.broker.model.DeprovisionRequest
+import com.swisscom.cloud.sb.broker.model.ProvisionRequest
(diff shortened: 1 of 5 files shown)


broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/relationaldb/RelationalDbFacade.groovy
```diff
@@ -0,0 +1,134 @@
...
+import groovy.util.logging.Slf4j
+import org.springframework.stereotype.Component
+
+@Component
+@Slf4j
+class RelationalDbFacade {
+    public static final String MAX_CONNECTIONS = "max_connections"
+
+    ProvisionResponse provision(RelationalDbClient client, ProvisionRequest request, String databasePrefix = "CFDB_") {
+        String database = createDatabaseName(request.serviceInstanceGuid, databasePrefix)
```
seilc1: I'd personally prefer "databaseName" over "database" to not confuse it with a dbConnection or similar

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/relationaldb/RelationalDbServiceProvider.groovy
```diff
@@ -18,138 +18,52 @@ package com.swisscom.cloud.sb.broker.services.relationaldb
...
-    public static final String MAX_CONNECTIONS = "max_connections"
     protected final RelationalDbClientFactory dbClientFactory
     protected final RelationalDbConfig dbConfig
 
+    RelationalDbFacade relationalDbFacade
+
     RelationalDbServiceProvider(RelationalDbConfig dbConfig, RelationalDbClientFactory dbClientFactory) {
         this.dbConfig = dbConfig
         this.dbClientFactory = dbClientFactory
+        this.relationalDbFacade = new RelationalDbFacade()
```
seilc1: Should be IMHO injected. if we have have to add some Autowired property to the facade it else may not be created correctly.

PR #139, opened 2018-08-20
Add BOSH CredHub

ampersand8:
**Problem**
For certain environments it can be necessary to support multiple CredHubs. For example when provisioning service instances with BOSH (with an integrated CredHub) and providing the credentials for a binding in the Cloud Foundry Runtime CredHub.

**Solution**
Support two CredHubs with OSB.

**Note**
With the suggested solution many CredHubs can be used similar to BoshCredHubTemplate.


ampersand8:
Addresses issue #140

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/credhub/DefaultCredHubConfig.groovy
```diff
@@ -0,0 +1,11 @@
+package com.swisscom.cloud.sb.broker.services.credhub
+
+import groovy.transform.CompileStatic
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+
+@CompileStatic
+@Configuration
+@ConfigurationProperties(prefix = "com.swisscom.cloud.sb.broker.credhub")
+class DefaultCredHubConfig implements CertificateConfig {
```
meip: Can you put in-file comment what this config is used for compared to `CredHubConfig.groovy`

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/credhub/CredHubConfig.groovy
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.swisscom.cloud.sb.broker.services.credhub
+
+import com.swisscom.cloud.sb.broker.config.Config
+import groovy.transform.CompileStatic
+import org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordResourceDetails
+
+@CompileStatic
+trait CredHubConfig implements Config {
```
meip: Can you put in-file comment what this config is used for compared to `DefaultCredHubConfig.groovy`

PR #147, opened 2018-08-29
Replace slf4j-test with log4j2 with default JSON config.

seilc1:


broker/src/test/groovy/com/swisscom/cloud/sb/broker/util/LoggingRequestInterceptorSpec.groovy
```diff
@@ -15,49 +15,20 @@
...
-    TestLogger logger
-    String testURL = "https://developer.swisscom.com"
+    Logger logger
 
     def setup() {
         restTemplate = new RestTemplateBuilder().build()
-        logger = TestLoggerFactory.getTestLogger(LoggingRequestInterceptor.class)
-    }
-
-    def "happy path: check if RestTemplate GET request is logged"() {
```
ampersand8: Is it not possible to test those use cases?

seilc1: slf4j will complain about 2 output libraries:

SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/taasech3/.gradle/caches/modules-2/files-2.1/uk.org.lidalia/slf4j-test/1.2.0/2f5f40ac2f912f215334a1c9d2f52bbc378eda0d/slf4j-test-1.2.0.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/taasech3/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-slf4j-impl/2.7/382b070836b8940a02d28c936974db95e9bfc3a4/log4j-slf4j-impl-2.7.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [uk.org.lidalia.slf4jtest.TestLoggerFactory]

PR #148, opened 2018-08-30
Osb metrics to dedicated db

melanielc:
Change name of db in application.yml to name of dedicated osb metrics db as declared in the influxdb-ha-boshrelease jobs template.

### DIFF
--- a/broker/src/main/resources/application.yml
+++ b/broker/src/main/resources/application.yml
@@ -28,6 +28,7 @@ management:
   metrics:
     export:
       influx:
+        enabled: false
         uri: http://localhost:8086
         step: 5s
         userName:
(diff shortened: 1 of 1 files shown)


meip:
Do we have to adopt the new InfluxDB database name to `docker-compose`?
https://github.com/swisscom/open-service-broker/blob/5c7d5833c8aaf4d71a0ae9d57b075b2d83dd861f/broker/src/functional-test/resources/test-docker-compose.yml#L24-L27
You could add a new database tot the container with env variable `INFLUXDB_DB`.

melanielc:
I don't think so since according to their documentation the database will be created automatically if it does not exist yet, no need to provide its name

> management.metrics.export.influx.auto-create-db=true # Whether to create the Influx database if it does not exist before attempting to publish metrics to it.

PR #154, opened 2018-09-06
Add platform context to UpdateRequest

Lafunamor:
This feature adds a context object to the update request as it is present in the provision request.
This allows for context specific actions during update and it allows reuse of provision code sequences


meip:
A service instance may belong to a given context at provision time. Since we now also track `context` through a _update request_ I also propose to update the context for the service instance then.

Issue #143, opened 2018-08-23
Support shield v8


ampersand8:
# Problem
Shield v7 is no longer actively maintained and the v8 API is not backwards compatible. OSB should ideally support both versions transparently to allow independent updates of OSB and shield.

# Requirements
- [ ] Transparently support shield v7 and v8 API

# Proposal
- Extend ShieldClient to support both shield API versions.

# Links
- [Shield v8 API doc](https://github.com/starkandwayne/shield/blob/v8.0.15/docs/API.md)
- [Shield v7 API doc](https://github.com/starkandwayne/shield/blob/v0.10.9/docs/api/http.md)

PR #155, opened 2018-09-10
Support shield v8

ampersand8:
Support shield API v1 and v2.

See issue #143


broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientFactory.groovy
```diff
@@ -13,17 +13,26 @@
...
+    private List<ShieldRestClient> shieldRestClients
+
+    @Autowired
+    ShieldRestClientFactory(List<ShieldRestClient> shieldRestClients) {
+        this.shieldRestClients = shieldRestClients
+    }
+
+    ShieldRestClient build() {
+        for (ShieldRestClient shieldRestClient in shieldRestClients) {
+            if (shieldRestClient.matchVersion()) return shieldRestClient
```
meip: Is the list `shieldRestClients` in a particular order or (in a very edge case) if multiple would match it's not deterministic?

ampersand8: No it's by design not deterministic. The idea is to simply add a class ShieldRestClientv3 and not having to define it anywhere else. I guess we could add a priority field, and if multiple matches are found the build method would return the implementation with the highest priority. I didn't put that in because it adds complexity for something that doesn't seem to be a requirement. What do you think?

PR #186, opened 2018-10-30
Feature/test helper service

melanielc:


broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/BaseFunctionalSpec.groovy
```diff
@@ -43,6 +45,9 @@ abstract class BaseFunctionalSpec extends Specification {
     @Autowired
     protected WebApplicationContext applicationContext
 
+    @Autowired
+    TestHelperService testHelperService
```
seilc1:
```suggestion
    protected TestHelperService testUtil
```

should only be used from within the extending class.

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/BaseFunctionalSpec.groovy
```diff
@@ -74,6 +79,14 @@ abstract class BaseFunctionalSpec extends Specification {
         }
     }
 
+    void activateServicesAndPlans(List<CFService> cfServices, List<Plan> plans) {
```
seilc1: Remove. The Coder can go directly via the `service`. there isn't any benefit with also providing these functions.

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/TestHelperServiceSpec.groovy
```diff
@@ -0,0 +1,115 @@
...
+        serviceBrokerClientExtended = createServiceBrokerClient()
+        def serviceId = "serviceDefinitionWithInstance"
+        def planId = "planForServiceDefinitionFromServiceManifestWithInstance"
+
+        and:
+        serviceBrokerClientExtended.createOrUpdateServiceDefinition(Resource.readTestFileContent("/service-data/serviceDefinitionWithInstance.json"))
+        assert (cfServiceRepository.findByGuid(serviceId))
+        assert (planRepository.findByGuid(planId))
+
+        and:
```
seilc1: why create a serviceInstance (which isn't deleted afterwards)?

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/TestHelperServiceSpec.groovy
```diff
@@ -0,0 +1,115 @@
...
```
seilc1: same as above?

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/TestHelperServiceSpec.groovy
melanielc: It needs to be created in order for the service definition to be set to inactive rather than just deleted because it is not used. But I will add a cleanup to delete it.

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/TestHelperServiceSpec.groovy
melanielc: same as above

PR #190, opened 2018-11-01
Fixes bug where shield backups could not be deleted

Lafunamor:
Fixes bug where shield backups could not be deleted due to a 404 returned by shield


MatthiasWinzeler:
@Lafunamor do we have an idea why the backups got deleted from shield in the first place? I saw the same error, but I couldn't figure out why it was deleted out of band. this fix looks more only working around the symptoms than to find the root cause.

PR #196, opened 2018-11-14
Feature/add auditlog

seilc1:


broker/src/main/groovy/com/swisscom/cloud/sb/broker/controller/ProvisioningController.groovy
```diff
@@ -81,32 +82,49 @@ class ProvisioningController extends BaseController {
...
-            return new ResponseEntity<ProvisionResponseDto>(new ProvisionResponseDto(dashboard_url: provisionResponse.dashboardURL),
-                    provisionResponse.isAsync ? HttpStatus.ACCEPTED : HttpStatus.CREATED)
+        def failed = false
+        def hasSensitiveData = false 
+        
+        try {
+            failIfServiceInstanceAlreadyExists(serviceInstanceGuid)
+
+            def serviceProvider = serviceProviderLookup.findServiceProvider(getAndCheckPlan(provisioningDto.plan_id))
+            if(serviceProvider instanceof SensitiveParameterProvider){
```
ampersand8: Is there a specific reason, that this is not handled similar to UpdateRequests where you added the method `boolean hasSensitiveParameters(Plan plan)` to UpdatingService?

PR #200, opened 2018-11-19
Add Bearer Authorization for RestTemplateBuilder

simonkrenger:
This commit adds authorization via Bearer token for the
RestTemplateBuilder. The additional method will add an "Authorization"
HTTP header to any request via interceptors.

### DIFF
--- a/broker/src/integration-test/groovy/com/swisscom/cloud/sb/broker/util/RestTemplateBuilderTest.groovy
+++ b/broker/src/integration-test/groovy/com/swisscom/cloud/sb/broker/util/RestTemplateBuilderTest.groovy
@@ -70,6 +70,19 @@ class RestTemplateBuilderTest extends Specification {
         httpServer?.stop()
     }
 
+    def "restTemplate with Bearer Token"() {
+        given:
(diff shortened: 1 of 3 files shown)


meip:
@MatthiasWinzeler just give a valuable hint that `BoshRestClient` already implements a UAA-Auth with Bearer token:  
https://github.com/swisscom/open-service-broker/blob/4a20506bcaac1b9678435a20fdd60c656d37ffd1/broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/HttpHelper.groovy#L34-L47

So it would make sense to align the two approaches to one. Would it make sense to separate the UAA stuff from `BoshRestClient` into something like `UaaClient` or just `BoshRestClient` to use the new `RestTemplateBuilder` functionality? /cc @ampersand8 ?

ampersand8:
Yeah, I completely agree with @MatthiasWinzeler. I'd suggest that the `BoshRestClient` makes use of the `RestTemplateBuilder`. I guess we could simply add e.g. a `createAuthRestTemplate` which would return a restTemplate with either simple or bearer authentication. With that we can get rid of the HttpHelper.

PR #203, opened 2018-12-03
enrich logs with service instance id

ampersand8:
Adds the service instance id for all log entries (requests and abstract jobs).

**Example:**
`2018-12-03 10:57:55 875774c7-bcd2-4141-83d3-37299e4dfcfc ERROR Test`

**Limitation:**
If the service instance id is used as a `@PathVariable`, you have to call it `serviceInstanceGuid`, as this is the name used in the interceptor.


broker/src/main/resources/log4j2.properties
```diff
@@ -19,7 +19,7 @@ name = PropertiesConfig
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
+appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss} %X{serviceInstanceGuid} %-5p %c{1}:%L - %m%n
```
seilc1: I'd suggest to move the serviceInstanceGuid to the end of the line as it is only used to filter ... and is not really interesting for reading the log ... ergo it will be annoying to always see the guid first, also log lines will not align with those not having a serviceInstanceGuid.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/async/job/AbstractLastOperationJob.groovy
```diff
@@ -48,6 +49,7 @@ abstract class AbstractLastOperationJob extends AbstractJob {
     @Override
     void execute(JobExecutionContext jobExecutionContext) throws JobExecutionException {
         String id = getJobId(jobExecutionContext)
+        MDC.put("serviceInstanceGuid", id)
```
seilc1: id is the job id not the service instance guid.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/job/AbstractBackupRestoreJob.groovy
```diff
@@ -40,6 +41,7 @@ abstract class AbstractBackupRestoreJob<T> extends AbstractJob {
     @Override
     void execute(JobExecutionContext jobExecutionContext) throws JobExecutionException {
         String id = getJobId(jobExecutionContext)
+        MDC.put("serviceInstanceGuid", id)
```
seilc1: as above

broker/src/main/groovy/com/swisscom/cloud/sb/broker/async/job/AbstractLastOperationJob.groovy
ampersand8: It is the service instance guid for provisioning, deprovisioning and update jobs.

ProvisioningJobConfig: [pass service instance guid as guid](https://github.com/swisscom/open-service-broker/blob/feature/enrich_log_context/broker/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/job/ProvisioningJobConfig.groovy#L29)

DeprovisioningJobConfig: [pass service instance guid as guid](https://github.com/swisscom/open-service-broker/blob/feature/enrich_log_context/broker/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/job/DeprovisioningJobConfig.groovy#L28)

UpdateJobConfig: [pass service instance guid as guid](https://github.com/swisscom/open-service-broker/blob/feature/enrich_log_context/broker/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/job/UpdateJobConfig.groovy#L28)

JobConfig: [assign guid](https://github.com/swisscom/open-service-broker/blob/feature/enrich_log_context/broker/src/main/groovy/com/swisscom/cloud/sb/broker/async/job/JobConfig.groovy#L39)

JobManager: [assign JobConfig.guid as Job id](https://github.com/swisscom/open-service-broker/blob/feature/enrich_log_context/broker/src/main/groovy/com/swisscom/cloud/sb/broker/async/job/JobManager.groovy#L48-L57)

broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/job/AbstractBackupRestoreJob.groovy
ampersand8: You're right, thanks for the catch.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/async/job/AbstractLastOperationJob.groovy
seilc1: oukay - true :-)

broker/src/main/resources/log4j2.properties
seilc1: mhm depends on the use case ... I was thinking on the usecase:

* when looking for a failure on a specific service isntance - you'll filter for this service instance - so seeing the serviceinstanceguid is not interesting anymore, because all lines have the same value there. - that's my reasoning to move it to the end.

in a common setup where you just scroll through the log it's interesting to see about which service instance the current log entry is.

Now I wonder which is the more relevant use case.

PR #205, opened 2018-12-05
set the serviceinstanceguid as log context

ampersand8:
The job id for backup and restore jobs is not the service instance guid. With this fix, the correct log context is being set with MDC.


broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/job/AbstractBackupRestoreJob.groovy
```diff
@@ -41,14 +41,15 @@ abstract class AbstractBackupRestoreJob<T> extends AbstractJob {
     @Override
     void execute(JobExecutionContext jobExecutionContext) throws JobExecutionException {
         String id = getJobId(jobExecutionContext)
-        MDC.put("serviceInstanceGuid", id)
+        MDC.put("serviceInstanceGuid", getServiceInstanceGuid(id))
```
seilc1: the value could be null if the backup/restore does not exist ... it's not a given that a null value is supported ... I suggest:

getServiceInstanceGuid(id) ?: "unknown"

PR #208, opened 2018-12-14
Catch LogContextEnrichIntercpetor exceptions

ampersand8:
Ensure all LogContextEnrichInterceptor exceptions are caught, as this should not interfere with operations.


ampersand8:
You mean something like `00000000-0000-0000-0000-000000000000`?

PR #216, opened 2019-01-10
Feature/backup cleanup

seilc1:


meip:
Would you elaborate the motivation to manage the backup retention from OSB instead of the backup system (i.e. Shield)?

ampersand8:
@meip To be precise this cleanup is about cleaning up the database, as we do for service instances.  Currently we only mark backups as deleted, but we never clean them up afterwards.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/backup/BackupService.groovy
```diff
@@ -136,8 +136,13 @@ class BackupService {
 
     private void checkIfAlreadyMarkedForDeletion(Backup backup) {
         if (Backup.Operation.DELETE == backup.operation) {
-            log.info("Backup with:${backup.guid} is already marked for deletion")
-            return
+            if (Backup.Status == Backup.Status.SUCCESS) {
+                log.info("Backup with:${backup.guid} is already deleted (awaiting cleanup job)")
+                ErrorCode.BACKUP_NOT_FOUND.throwNew()
```
ampersand8: @seilc1 As @meip suggested to me, we maybe should return a 410 GONE. And looking at the [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) I think it fits this exact scenario better, as this indicates to clients the backup existed and was intentionally removed.

PR #219, opened 2019-01-28
ParentServiceProvider to handle parent service providers

ampersand8:
**Problem:**
Services provided in a parent / child relationship have to bring their own logic to manage the given restrictions. This leads to a lot of duplicated code, as the problems are bound to be very similar.

**Proposal:**
The trait ParentServiceProvider will provide the default logic, which is automatically applied when the ServiceProvider `implements ParentServiceProvider`. The default logic can be overridden by every ServiceProvider.

**Limitations:**
In the plan for the parent serviceprovider the parameter `max_children` has to be set (default is unlimited).

### DIFF
--- a/broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
+++ b/broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
@@ -15,7 +15,7 @@ trait ParentServiceProvider {
         if (param == null) {
             return false
         } else {
-            return (param.value as int) >= serviceInstance.childs.count({ !it.deleted })
+            return (param.value as int) <= serviceInstance.childs.count({ !it.deleted })
         }
     }
 }
--- a/broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
+++ b/broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
@@ -1,0 +1,16 @@
+package com.swisscom.cloud.sb.broker.util
+
+import com.swisscom.cloud.sb.broker.model.ServiceInstance
+
+trait ParentServiceProvider {
+    private static final String MAX_CHILDREN = "max_children"
+
+    boolean hasActiveChildren(ServiceInstance serviceInstance) {
+        return serviceInstance.childs.any({!it.deleted})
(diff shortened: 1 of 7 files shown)


broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
```diff
@@ -0,0 +1,21 @@
...
+    boolean hasActiveChildren(ServiceInstance serviceInstance) {
+        return serviceInstance.childs.any({ !it.deleted })
+    }
+
+    boolean isFull(ServiceInstance serviceInstance) {
+        Parameter param = serviceInstance.plan.parameters.find { it.name == MAX_CHILDREN }
+        if (param == null) {
+            return false
+        } else {
+            return (param.value as int) <= serviceInstance.childs.count({ !it.deleted })
```
seilc1: missing a check for failed completion:

it.completed && !it.deleted

seilc1:
Intention is good, but I think we have to look at it in a bigger picture. We are lacking in general in the area of input validation ... I've been thinking about 3 additional functions on the ServiceProvider:

validateProvisionRequest()
validateDeprovisionRequest()
validateUpdateRequest()

this would allow us to give immediate/synchronous feedback if the request we got will be processed. This is not fool-proof as some of the checks have to be done by the backend (e.g. is there enough space). but we already can improve our service quality alot with that.
Validating the Parent/Child References is something which then also belongs for me into exactly this functions - or maybe not.

About validating the Parent-Child-References, I see at least the following things we should validate for Parent-Child-References:

- Does the parent exist (completed and !deleted)
- Does parent and child match together (e.g. I don't attach a MariaDB to MongoDBCluster || could also be something like can I deploy a unencrypted MariaDB to a encrypted MariaDB Cluster)
- Does the parent have enough space for another child (as done in this example ... or maybe more complex)

I would suggest to have a more detailed discussion about this.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
ampersand8: Hm, what if the child provisioning takes a while and the customer provisions a bunch of them. Since those requests would not yet be completed he'd be able to provision more than we set the limit to.

PR #222, opened 2019-03-04
WIP: remove unused functionality for bosh based services

ampersand8:
**Description:**
The functionality to create OpenStack server groups to specify affinity is no longer in use, so this can be removed. Furthermore setting the cloud-config, which was used in conjunction, can also be removed and will be replaced with [generic configs](https://bosh.io/docs/director-api-v1/#configs) to support future use cases.


ampersand8:
Because this is a breaking change, I agree this should be a major release.

PR #225, opened 2019-03-07
Chore/updating and cleaning gradle

ibonUrrutia:
I cleaned some smells found in the gradle build:
- `settings.gradle` should only appear in the root of a multimodule project
- I updated to gradle 5.2.1 to benefit from parallel builds and improvements in build speed
- Centralize all plugin definition on the root for updating versions easily
- Add dependency management using `gradle/dependencies.gradle`

WIP:
- Use maven-publish gradle plugin for uploading artifacts to repository


build.gradle
```diff
@@ -36,13 +20,32 @@ nexusStaging {
...
     apply plugin: 'signing'
 
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+
+    repositories {
+        jcenter()
+        mavenCentral()
+        maven {
+            url "http://repo.spring.io/libs-milestone-local"
```
seilc1: this shouldn't be needed anymore. the package has been updated to the real release.

ibonUrrutia:
> has the client test been removed? why?

I mentioned in the commit message: while I was updating the client to use the spring boot gradle plugin, it started to fail because spring boot was not finding some classes in the classpath. 
The deleted test was using a dummy broker using `@SpringBootApplication` and running that fake application from a `@BeforeClass` in the test. IMHO, it is better to use mocking and/or stubbing than that approach, for not having to deal with classpath issues in the tests or having to add dependencies in the tests only for the dummy broker.
In any case, we deleted the test because we were not understanding what was really testing and waited to somebody raising the question if it was important. Are you using it? We can revert the change but then discuss what is really trying to test.

seilc1:
the test is verifying that we can use our client implementation to talk to a valid (spring boot) open service broker implementation. it's checking that we use the correct paths and objects. and it also checks that we are using basic authentication in the client correctly.

ibonUrrutia:
Ok, I will rebase and revert my changes so the test is not deleted

PR #233, opened 2019-03-26
Refactor to separate broker-core from broker

meip:
**Problem:**
Service Broker module `broker` is a big chunk of code which implements the OSB specification on one hand, as well as implement plenty of service providers wich are very specific. 

**Proposal:**
Introducing a new module `broker-core` contains only the relevant stuff to serve the OSB specification while module `broker` is depending on that module and adds the service providers to it. If a project now doesn't require any of the given service providers it can reuse core while the `broker` module stays unchanged.


meip:
@ibonUrrutia I agree that import changes (to wildcard imports and back) are annoying for reviews. Since I don't see any benefits for wildcards import I would argue to try to avoid it for the future.

PR #236, opened 2019-04-10
Bugfix/null context on update deletes context

seilc1:
make sure that the context doesn't get deleted when updating a service instance.

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/context/ServiceContextPersistenceService.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/context/ServiceContextPersistenceService.groovy
@@ -79,7 +79,7 @@ class ServiceContextPersistenceService {
     }
 
     private void setServiceContextDetails(Map<String, Object> newDetails, ServiceContext serviceContext) {
-        def existingDetails = serviceContext.getDetails().toList()
+        List<ServiceContextDetail> existingDetails = serviceContext.getDetails().toList() as List<ServiceContextDetail>
 
         newDetails.each {
             newDetail ->
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
@@ -54,7 +54,9 @@ class UpdatingPersistenceService {
     void updatePlanAndServiceDetails(final ServiceInstance serviceInstance, final String updateParameters, final Collection<ServiceDetail> serviceDetails, Plan plan, ServiceContext serviceContext) {
         serviceInstance.plan = plan
         serviceInstance.parameters = mergeServiceInstanceParameter(serviceInstance.parameters, updateParameters)
-        serviceInstance.serviceContext = serviceContext
+        if (serviceContext) {
+            serviceInstance.serviceContext = serviceContext
+        }
         serviceInstanceRepository.saveAndFlush(serviceInstance)
         updateServiceDetails(serviceDetails, serviceInstance)
     }
--- a/broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceContextFunctionalSpec.groovy
+++ b/broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceContextFunctionalSpec.groovy
@@ -126,16 +126,12 @@ class ServiceContextFunctionalSpec extends BaseFunctionalSpec {
                 "field_c": 13.37,
                 "field_d": 1337
         ])
-
-        when:
         def serviceInstanceGuid = requestServiceProvisioning(new HashMap<String, Object>(),
                 defaultPlanGuid,
                 false,
                 defaultServiceGuid,
                 context)
 
-        then:
-        noExceptionThrown()
         def serviceInstance = serviceInstanceRepository.findByGuid(serviceInstanceGuid)
         serviceInstance.serviceContext.platform == "CUSTOM"
         serviceInstance.serviceContext.details.find { d -> d.key == "field_a" }.value == "Some Value"
@@ -145,7 +141,10 @@ class ServiceContextFunctionalSpec extends BaseFunctionalSpec {
 
         when:
         context = null
-        serviceLifeCycler.requestUpdateServiceInstance(serviceInstanceGuid, defaultServiceGuid, defaultPlanGuid, null, false, context)
+        serviceLifeCycler.requestUpdateServiceInstance(serviceInstanceGuid, defaultServiceGuid, defaultPlanGuid,
+                null,
+                false,
+                context)
 
         then:
         def serviceInstance2 = serviceInstanceRepository.findByGuid(serviceInstanceGuid)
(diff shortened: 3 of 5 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/context/ServiceContextPersistenceService.groovy
```diff
@@ -79,7 +79,7 @@ class ServiceContextPersistenceService {
     }
 
     private void setServiceContextDetails(Map<String, Object> newDetails, ServiceContext serviceContext) {
-        def existingDetails = serviceContext.getDetails().toList()
+        List<ServiceContextDetail> existingDetails = serviceContext.getDetails().toList() as List<ServiceContextDetail>
```
ibonUrrutia: Another motive for getting rid of groovy and do it with Java.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
```diff
@@ -54,7 +54,9 @@ class UpdatingPersistenceService {
     void updatePlanAndServiceDetails(final ServiceInstance serviceInstance, final String updateParameters, final Collection<ServiceDetail> serviceDetails, Plan plan, ServiceContext serviceContext) {
         serviceInstance.plan = plan
         serviceInstance.parameters = mergeServiceInstanceParameter(serviceInstance.parameters, updateParameters)
-        serviceInstance.serviceContext = serviceContext
+        if (serviceContext) {
+            serviceInstance.serviceContext = serviceContext
```
ibonUrrutia: Can be written as:
`serviceInstance.serviceContext = serviceContext` 
If ServiceContext.NULL is created: 
`public static final ServiceContext NULL = new ServiceContext();  //Instantiate with empty values` 

(that is called the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) plus syntactic sugar).

That way the code is not full of `if` checking for NULL, and there will be no `NullPointerException`

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceContextFunctionalSpec.groovy
```diff
@@ -117,4 +117,46 @@ class ServiceContextFunctionalSpec extends BaseFunctionalSpec {
         cleanup:
         serviceLifeCycler.deleteServiceInstanceAndAssert(serviceInstanceGuid, defaultServiceGuid, defaultPlanGuid, null, false, DummyServiceProvider.RETRY_INTERVAL_IN_SECONDS * 4)
     }
+
+    def "sending no context does not delete context"() {
```
ibonUrrutia: Don't like this all in one test. In this test what you REALLY are testing is:
- You can request a service provisioning (assert on line 138-144, the findByGuid should be on `when:` block)
- You can update a service with `null, false, null` and the service does not change (dit it actually has been updated? how do you know that the cal was really done?)

In spock usually the thing would be having a test for the first part, and a second test for the rest. In the second test you put in the given block ALL the initialisation code. 

Also, we should assert that the service was called by `requestUpdateServiceInstance` or the rest of the asserts does not make sense.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
seilc1: in this case not relevant, as it actually is an override, so it currently overrode a valid context with an empty one (which is not bad ... because the api-user just didn't wanna do anything).

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceContextFunctionalSpec.groovy
seilc1: There are tests for the first part in other classes, testing that simple provisioning works. 
The testing of updating a context requires a service to be there, and I prefer to have tests which are not dependent on other tests (else you need to run those in the correct order which might be difficult).

About moving the setup to the given I agree and I changed it.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
ibonUrrutia: How do you know that the client of the API don't  want to override the context but he forgot to initialize the context?
Imagine this client code:
````
Context newContext;   // Let's initialize it later
...
...
...
// Lots of lines of code, or better, lot of calls to methods chaining and in the last method:
updatePlanAndServiceDetails(serviceInstance, updateParameters, serviceDetails, plan, context)
```
If you follow my indications about `null` the method launches and `IllegalArgumentException`, the client of the API discovers he had forgotten and corrects it. If he doesn't want to update the context he passes this:
````
updatePlanAndServiceDetails(serviceInstance, updateParameters, serviceDetails, plan, EMPTY)
```
which IMHO is much more clear about his intention.

In your case, the `null` could mean:
- A mistake in the code
- I don't want to update the context
Which is usually what happens with `null`: it means different things.

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/ServiceContextFunctionalSpec.groovy
ibonUrrutia: I didn't ask to depend on other test: we should NOT depend on other tests. I was asking to show in the test what is `given` (assumed to work), what are we testing (`when`), and what we are expecting (`then`)... and if is the case, and here I think that is the case, test the `when` with different `where` ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/updating/UpdatingPersistenceService.groovy
seilc1: This object is generated based on the values handed over the REST Controller ... so there is a "null" in the serviceContext Property created by the SpringBoot RestController ... which is 'valid' I can't help the user of the API ... as this could have been done via curl. 

as you see here: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#updating-a-service-instance it specifies the context as not mandatory.

ibonUrrutia: Umm, sorry but I don't  understand you:
- If the object is optional , you define a Jackson Serializer that converts `EMPTY` to nothing, or `{}`, or `{ "name":""...}` , depending in what is understood by the API/mandated
- If the method can receive nulls:
``context = context == null ? EMPTY:context`` and then you continue normally

seilc1: Oukay ... I see ... sounds like a good solution ... but will require bigger changes spaning the whole Controller logic. IMHO we can put that in the refactor for switching to spring boot service broker. there we have anyway to touch everything.

PR #237, opened 2019-04-10
fix bosh config deletion by ignoring not found exception and setting correct url query

ampersand8:
The url query for the bosh config API has to be carefully set, the setFilter helper makes sure the query is valid. With this change the helper is also used when executing a bosh delete config call.

Furthermore during the deletion of the bosh config the exception `BoshResourceNotFound` should be ignored, as it's possible that there is no bosh config at the time of deprovisioning.

Moving the handling of `BoshResourceNotFound` to the BoshRestClient makes handling of the `deleteDeployment` and `deleteBoshConfig` homogenous.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -144,16 +144,16 @@ class BoshRestClient {
...
     String getConfigs(String name, String type) {
-        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type)), HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
+        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type, true)), HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
     }
 
-    private String setFilter(String name, String type) {
+    private String setFilter(String name, String type, Boolean latest = null) {
         URIBuilder uriBuilder = new URIBuilder()
-        uriBuilder.setParameter('latest', 'true')
+        if (latest != null) uriBuilder.setParameter('latest', latest.toString())
```
seilc1: what happens at latest false? - does it take a random one, or the second to last?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshClient.groovy
```diff
@@ -105,7 +105,12 @@ class BoshClient {
         new JsonSlurper().setType(RELAX).parseText(boshRestClient.getConfigs(name, type)) as List<BoshConfigResponseDto>
     }
 
-    void deleteConfig(String name, String type) {
-        boshRestClient.deleteConfig(name, type)
+    void deleteConfigIfExists(String name, String type) {
+        try {
+            boshRestClient.deleteConfig(name, type)
+        } catch (BoshResourceNotFoundException e) {
+            log.debug("Bosh Config with name ${name} and type ${type} not found. This is ignored during deletion.")
```
ibonUrrutia: Should launch `IllegalArgumentException` : if somebody makes a mistake, you are warning him. If is not a mistake he could catch your exception and then log.error. But if you are doing this, when log level is info, is the same as doing an empty catch, and we don't do empty catch ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ibonUrrutia: why this conversation is resolved?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -144,16 +144,16 @@ class BoshRestClient {
...
+        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type)), HttpMethod.DELETE, new HttpEntity<Object>(createAuthHeaders()), String.class)
     }
 
     String getConfigs(String name, String type) {
-        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type)), HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
+        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type, true)), HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
     }
 
-    private String setFilter(String name, String type) {
+    private String setFilter(String name, String type, Boolean latest = null) {
```
ibonUrrutia: Why `Boolean` instead of `boolean`? If it is for allowing `null`, you know my opinion.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClientSpec.groovy
```diff
@@ -196,7 +196,7 @@ class BoshRestClientSpec extends Specification {
         String name = 'test'
         String type = 'cloud'
         when:
-        String filter = boshRestClient.setFilter(name, type)
+        String filter = boshRestClient.setFilter(name, type, true)
```
ibonUrrutia: umm, setFilter is private... why testing it? and specially , how is that allowed?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshConfigAPIQueryFilterParameter.groovy
```diff
@@ -0,0 +1,20 @@
+package com.swisscom.cloud.sb.broker.services.bosh.client
+
+enum BoshConfigAPIQueryFilterParameter {
```
ibonUrrutia: All API must have javadoc explaining what and specially why.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -144,18 +144,23 @@ class BoshRestClient {
     }
 
     void deleteConfig(String name, String type) {
-        createRestTemplate().exchange(prependBaseUrl("${CONFIGS}?name=${name}&type=${type}"), HttpMethod.DELETE, new HttpEntity<Object>(createAuthHeaders()), String.class)
+        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter([(BoshConfigAPIQueryFilterParameter.NAME): name,
```
ibonUrrutia: Use static imports for making more readable

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -144,18 +144,23 @@ class BoshRestClient {
...
     String getConfigs(String name, String type) {
-        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter(name, type)), HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
+        createRestTemplate().exchange(prependBaseUrl(CONFIGS + setFilter([(BoshConfigAPIQueryFilterParameter.NAME)  : name,
+                                                                          (BoshConfigAPIQueryFilterParameter.TYPE)  : type,
+                                                                          (BoshConfigAPIQueryFilterParameter.LATEST): 'true'])),
+                HttpMethod.GET, new HttpEntity(createAuthHeaders()), String.class).body
     }
 
-    private String setFilter(String name, String type) {
+    private String setFilter(Map<BoshConfigAPIQueryFilterParameter, String> filter) {
```
ibonUrrutia: Instead of passing a Map, why not create a `BoshConfigAPIQueryFilterParameters` Immutable class. You could have methods like :
``` 
setFilter(BoshConfigAPIQueryFilterParameters.of(name, type, true))
```
or 
``` 
setFilter(BoshConfigAPIQueryFilterParameters.of(name, type))
```
And:
```
BoshConfigAPIQueryFilterParameters.asUriString()
```
Which will make your API more readable, easier to use and less error prone.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacadeSpec.groovy
```diff
@@ -241,7 +241,7 @@ class BoshFacadeSpec extends Specification {
         boshFacade.deleteConfig('test', 'cloud')
 
         then:
-        1 * boshClient.deleteConfig(_, _)
+        1 * boshClient.deleteConfigIfExists(_, _)
```
ibonUrrutia: In stubs I prefer to pass the exact values in test. You know, sometimes the test fails really, but because is any parameter, it passes nulls :-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshClient.groovy
seilc1: disagree: you try to delete something which doesn't exist anymore ... this could be remedy or cleanup ... if the thing you want to delete is already gone ... everything is fine.

ampersand8: `BoshResourceNotFoundException` could be triggered if either, the search parameters are false or that there is actually no bosh config for this deployment.

So e.g. if in the deployment phase there was no bosh config set [here](https://github.com/swisscom/open-service-broker/blob/develop/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy#L93), because the deployment may not need it, there would be no config to delete. With this permissive deletion we risk having configs around that should have been deleted.

To mitigate that risk we could delete it the same way it was created, maybe something like this [in BoshFacade](https://github.com/swisscom/open-service-broker/blob/develop/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy#L160) and not catch the exception:
```
for (config in serviceConfig.boshConfigs) {
    deleteConfig(lastOperationJobContext.serviceInstance.guid, config.type)
}
```

I agree with the log level anyway, should inform about the catch in any case.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: Yeah, it was for `null`. Changed it now.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: Refactored now to no longer use `setFilter` and use the `GenericConfigAPIQueryFilter` class instead (with no Maps in sight).

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: Moved this to `GenericConfigAPIQueryFilter`.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacadeSpec.groovy
ampersand8: You're right there's no reason here to use the any parameter. Changed to specific parameters now.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshConfigAPIQueryFilterParameter.groovy
ampersand8: Hope [that](https://github.com/swisscom/open-service-broker/blob/fix/set_bosh_configs_urlquery_with_helper/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/GenericConfigAPIQueryFilter.groovy#L20) explanation is enough doc.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClientSpec.groovy
ampersand8: Removed `setFilter`.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshClient.groovy
ampersand8: implemented my suggested solution [here](https://github.com/swisscom/open-service-broker/blob/fix/set_bosh_configs_urlquery_with_helper/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy#L159-L161).

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: The question was about the `latest` parameter, this should be answered by the reference to the [bosh director HTTP API doc](https://github.com/swisscom/open-service-broker/blob/fix/set_bosh_configs_urlquery_with_helper/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/GenericConfigAPIQueryFilter.groovy#L22).

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshConfigAPIQueryFilterParameterSpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package com.swisscom.cloud.sb.broker.services.bosh.client
+
+import spock.lang.Specification
+
+class BoshConfigAPIQueryFilterParameterSpec extends Specification {
+    def "happy path: query correctly built with name and type set"() {
```
ibonUrrutia: I prefer to call the test methods "should ..." or "should not ... "
Happy and bad path is developers slang, and sometimes hard to understand for non-devs (uh? is there a sad part on our product/project?).

PR #238, opened 2019-04-10
Bugfix/failed serviceinstances still count as active children

seilc1:
Make sure to only count created or creating children as active children and not children which have crashed during provision.

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
@@ -26,10 +26,10 @@ trait ParentServiceProvider {
     }
 
     int getActiveChildrenCount(ServiceInstance serviceInstance) {
-        List<ServiceInstance> undeletedChildren = serviceInstance.childs.findAll { si -> !si.deleted }
+        def undeletedChildren = serviceInstance.childs.findAll { si -> !si.deleted }
         int completedUndeletedChildrenCount = undeletedChildren.count { si -> si.completed }
 
-        List<ServiceInstance> uncompletedChildren = undeletedChildren.findAll { si -> !si.completed }
+        def uncompletedChildren = undeletedChildren.findAll { si -> !si.completed }
 
         int childrenProvisionInProgress = 0
         uncompletedChildren.each {
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
@@ -1,13 +1,19 @@
 package com.swisscom.cloud.sb.broker.util
 
+import com.swisscom.cloud.sb.broker.model.LastOperation
 import com.swisscom.cloud.sb.broker.model.Parameter
 import com.swisscom.cloud.sb.broker.model.ServiceInstance
+import com.swisscom.cloud.sb.broker.model.repository.LastOperationRepository
+import org.springframework.beans.factory.annotation.Autowired
 
 trait ParentServiceProvider {
     private static final String MAX_CHILDREN = "max_children"
 
+    @Autowired(required = true)
+    LastOperationRepository lastOperationRepository
+
     boolean hasActiveChildren(ServiceInstance serviceInstance) {
-        return serviceInstance.childs.any({ !it.deleted })
+        return getActiveChildrenCount(serviceInstance) > 0
     }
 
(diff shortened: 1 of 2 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
```diff
@@ -1,21 +1,45 @@
...
+        def undeletedChildren = serviceInstance.childs.findAll { si -> !si.deleted }
+        int completedUndeletedChildrenCount = undeletedChildren.count { si -> si.completed }
+
+        def uncompletedChildren = undeletedChildren.findAll { si -> !si.completed }
+
+        int childrenProvisionInProgress = 0
+        uncompletedChildren.each {
+            si ->
+                def lastOperation = lastOperationRepository.findByGuid(si.guid)
+                if (lastOperation.status == LastOperation.Status.IN_PROGRESS) {
```
ibonUrrutia: Use static imports, `lastOperation.status == IN_PROGRESS` reads much better

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/ParentServiceProvider.groovy
```diff
@@ -1,21 +1,45 @@
...
 
+    @Autowired(required = true)
+    LastOperationRepository lastOperationRepository
+
     boolean hasActiveChildren(ServiceInstance serviceInstance) {
-        return serviceInstance.childs.any({ !it.deleted })
+        return getActiveChildrenCount(serviceInstance) > 0
     }
 
     boolean isFull(ServiceInstance serviceInstance) {
```
ibonUrrutia: I'm not getting this `isFull` method:
- If there is not `MAX_CHILDREN` parameter, returns false. So, a service without that parameter is never full? What happens if there is no parameter by mistake?
I would prefer ``if(param == null && hasActiveChildren) return false`` or better:
```
boolean canBeFull(ServiceInstance serviceInstance){
   return  serviceInstance.plan.parameters.find { it.name == MAX_CHILDREN } != null // or > negative number
}
boolean isFull(ServiceInstance si){
if(!canBeFull(si)) { return false} else ...
```

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/provisioning/ProvisioningServiceSpec.groovy
```diff
@@ -175,6 +180,8 @@ class ProvisioningServiceSpec extends Specification {
     def "checkParent() fails with parent_reference to full parent service instance"() {
         given:
         def serviceProvider = Mock(ServiceProvider).withTraits(ParentServiceProvider)
+        serviceProvider.lastOperationRepository = lastOperationRepository
+        lastOperationRepository.findByGuid(_) >> new LastOperation(status: LastOperation.Status.IN_PROGRESS)
```
ibonUrrutia: I prefer to stub referring to certain value than passing any value, usually I'm using the value in test:
lastOperationRepository.findByGuid("guid in test")

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/provisioning/ProvisioningServiceSpec.groovy
```diff
@@ -205,6 +212,8 @@ class ProvisioningServiceSpec extends Specification {
     def "checkParent() succeeds with parent_reference to a parent service instance"() {
         given:
         def serviceProvider = Mock(ServiceProvider).withTraits(ParentServiceProvider)
+        serviceProvider.lastOperationRepository = lastOperationRepository
+        lastOperationRepository.findByGuid(_) >> new LastOperation(status: LastOperation.Status.IN_PROGRESS)
```
ibonUrrutia: Same as previous comment

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/provisioning/ProvisioningServiceSpec.groovy
```diff
@@ -235,6 +244,8 @@ class ProvisioningServiceSpec extends Specification {
     def "checkActiveChildren fails with active children"() {
         given:
         def serviceProvider = Mock(ServiceProvider).withTraits(ParentServiceProvider)
+        serviceProvider.lastOperationRepository = lastOperationRepository
+        lastOperationRepository.findByGuid(_) >> new LastOperation(status: LastOperation.Status.IN_PROGRESS)
```
ibonUrrutia: Same as previous

PR #239, opened 2019-04-12
chore: Using jfrog gradle plugin to publish

ibonUrrutia:
We are having problems for publishing SNAPSHOTS at oss.jfrog.org with
nebula plugin, so we are trying to do it with the jfrog plugin.
Unfortunately our bintray users have some problem and the SNAPSHOT is
not published because a HTPP 403 Forbidden. Support ticket created in
JFrog.


ibonUrrutia:
We managed to publish to bintray and maven central using updated versions of nebula publishing plugins, so we don't need this approach anymore.

PR #243, opened 2019-04-18
Feature/change release workflow

ibonUrrutia:
We are moving from our current release workflow (that is making `master` and `develop` diverge and not tagging correctly) to a workflow based purely in gradle plugins for [versioning](https://github.com/allegro/axion-release-plugin/) and [publishing](https://github.com/nebula-plugins/nebula-publishing-plugin).

Some of the benefits will be:
- Less noise in the commit history of the project
- Tags will be pointing to real commits in `develop`
- Try to unify the diverged history of `master` and `develop`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,26 +10,26 @@ plugins {
     id "com.github.jk1.dependency-license-report" version "1.2"
 }
 
-
+// According to axion documentation you should apply the scmVersion configuration in the root project and associate
+// the calculated version to subprojects
+// https://axion-release-plugin.readthedocs.io/en/latest/configuration/basic_usage/#multi-module-with-multiple-versions
 apply from: "$rootDir/gradle/versioning.gradle"
 
 allprojects {
     project.version = scmVersion.version
-
 }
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -5,27 +5,29 @@ plugins {
     id 'nebula.source-jar' version '10.0.0' apply false
     id 'nebula.test-jar' version '10.0.0' apply false
     id 'nebula.nebula-bintray' version '7.0.0' apply false
+    id 'pl.allegro.tech.build.axion-release' version '1.10.0'
     id 'org.springframework.boot' version '2.1.1.RELEASE' apply false
     id "com.github.jk1.dependency-license-report" version "1.2"
 }
 
 
+apply from: "$rootDir/gradle/versioning.gradle"
```
ampersand8: I'm not sure to why `versioning.gradle` is applied here, while `repositories.gradle`, `dependencies.gradle` and `functional-testing.gradle` are applied in subprojects. As discussed a comment should resolve this uncertainty.

PR #255, opened 2019-04-25
fix: Move c.s.c.s.b.b.c.enable to o.b.c.enable

ibonUrrutia:
Again, like in the case of osb.credhub.enable, if you use that enable
property in BoshCredHubTemplate, you are
first loading the credhub stuff and then enabling or disabling it. In
the case of disabling it does not make sense to be forced to configure
first.


ibonUrrutia:
Don't merge, just discovered that still doesn't solve the problem.

ibonUrrutia:
@ampersand8 please, review my changes.

broker/src/functional-test/groovy/com/swisscom/cloud/sb/broker/functional/BindingParametersFunctionalSpec.groovy
```diff
@@ -34,7 +34,7 @@ class BindingParametersFunctionalSpec extends BaseFunctionalSpec {
     @Autowired
     private ServiceBindingPersistenceService serviceBindingPersistenceService
     @Autowired
-    private CredHubCredentialStoreStrategy credHubCredentialStoreStrategy
+    private CredHubCredentialStore credHubCredentialStoreStrategy
```
ampersand8: I'd suggest to remove the `Strategy` in the variable name as well.

PR #256, opened 2019-04-25
chore: fix Gradle publishing for OSB developer README

meip:
fix parameter name `bintrayPassword` to `bintrayApiKey`


meip:
see #257

PR #261, opened 2019-05-03
refactor/convert credhub test to ci

ibonUrrutia:
We started using [Test Containers](https://www.testcontainers.org/) for launching a docker compose instance automatically when integration testing. Using [Junit Category]() we can differentiate the tests between unit and integration test.


ibonUrrutia:
I'm going to open a new branch for moving all integration-tests to testcontainers in a straightforward way: updating the base specification. It will be faster to deploy just one docker-compose and then launch all integration tests against it.

PR #262, opened 2019-05-10
test: add tests for BoshFacade with WireMock

ampersand8:
**Problem:**
Bosh related features are hard to test, as an according backend is required and there are no ideal solutions with docker available.

**Proposal:**
With well defined WireMock tests, continuous testing of the BoshFacade is very easily possible, as the tests are fast and have no requirements for external systems.

**Attention:**
To improve testability of the `BoshFacade`, many surrounding configs needed to be changed. Most notably `serviceTemplates` are now part of the specific services. This applies to k8s as well as bosh services.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -88,48 +84,15 @@ class BoshFacade {
...
 
-    void handleTemplatingAndCreateConfigs(ProvisionRequest provisionRequest, BoshTemplateCustomizer templateCustomizer) {
+    List<BoshConfigResponseDto> handleTemplatingAndCreateConfigs(ProvisionRequest provisionRequest, BoshTemplateCustomizer templateCustomizer) {
+        List<BoshConfigResponseDto> boshConfigResponseDtos = new ArrayList<>()
         for (config in serviceConfig.genericConfigs) {
             BoshTemplate template = boshTemplateFactory.build(templateConfig.getTemplateForServiceKey(config.templateName).first())
             template.replace(PARAM_GUID, provisionRequest.serviceInstanceGuid)
             templateCustomizer.customizeBoshConfigTemplate(template, config.type, provisionRequest)
-            createBoshClient().setConfig(new BoshConfigRequestDto(name: provisionRequest.serviceInstanceGuid, type: config.type, content: template.build()))
+            boshConfigResponseDtos.add(createBoshClient().setConfig(new BoshConfigRequestDto(name: provisionRequest.serviceInstanceGuid, type: config.type, content: template.build())))
```
ibonUrrutia: Consider adding a builder to BoshConfigRequestDto to avoid this constructor calls

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -88,48 +84,15 @@ class BoshFacade {
         return serviceDetails
     }
 
-    void handleTemplatingAndCreateConfigs(ProvisionRequest provisionRequest, BoshTemplateCustomizer templateCustomizer) {
+    List<BoshConfigResponseDto> handleTemplatingAndCreateConfigs(ProvisionRequest provisionRequest, BoshTemplateCustomizer templateCustomizer) {
+        List<BoshConfigResponseDto> boshConfigResponseDtos = new ArrayList<>()
```
ibonUrrutia: [IMHO] I usually call the variable that a method returns `result`: this way is always clear inside a method what is its objective to return and why is doing certain things to that variable

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -202,4 +153,57 @@ class BoshFacade {
         return boshTemplateFactory
     }
 
+    @VisibleForTesting
```
ibonUrrutia: Still I'm not getting this `@VisibleForTesting` annotations: if you use the java normal scopes `public`, `package` and `private`, you can test the API, the test is usually in the same package so you can test whatever you don't want to put in the API and testing things that nobody can call except the same class makes no sense for me.

In any case, testing this method makes no sense.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -202,4 +153,57 @@ class BoshFacade {
...
+    private static String generateDeploymentId(String serviceInstanceGuid) {
+        return DEPLOYMENT_PREFIX + serviceInstanceGuid
+    }
+
+    private static String findBoshTaskIdForDeploy(LastOperationJobContext context) {
+        return ServiceDetailsHelper.from(context.serviceInstance.details).getValue(BoshServiceDetailKey.BOSH_TASK_ID_FOR_DEPLOY)
+    }
+
+    @VisibleForTesting
```
ibonUrrutia: Same comment as before about `@VisibleForTesting`

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -202,4 +153,57 @@ class BoshFacade {
...
+    private List<String> generateHostNames(String guid, int hostCount) {
+        (List<String>) (0..<hostCount).inject([]) {
+            result, i -> result.add("${guid}-${i}${HOST_NAME_POSTFIX}"); return result
+        }
+    }
+
+    private String readTemplateContent(String templateIdentifier, String version = "1.0.0") {
+        Preconditions.checkNotNull(templateIdentifier, 'Need a valid template name!')
```
ibonUrrutia: If the variable is called `templateIDentifier`, change exception message to  `Invalid template identifier`. 

What is happening when `templateIdentifier` is empty? What is happening when `templateIdentifier` is ` `? More than checking in this private method null, assure that the class is not calling to this method with an invalid identifier.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -202,4 +153,57 @@ class BoshFacade {
...
+        provisionRequest.plan.parameters?.each { Parameter p -> template.replace(p.name, p.value) }
+    }
+
+
+    private BoshClient createBoshClient() {
+        return boshClientFactory.build(serviceConfig)
+    }
+
+    private static String findBoshDeploymentId(LastOperationJobContext context) {
+        return ServiceDetailsHelper.from(context.serviceInstance.details).findValue(BoshServiceDetailKey.BOSH_DEPLOYMENT_ID).or(generateDeploymentId(context.serviceInstance.guid))
```
ibonUrrutia: Format this line so I don't have to use scroll for reading it.

In general: Use `import static` when using static methods and each call to a method in a calling chain should be in its own line

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshClient.groovy
```diff
@@ -67,7 +67,7 @@ class BoshClient {
         try {
             return Optional.of(boshRestClient.deleteDeployment(id))
         } catch (BoshResourceNotFoundException e) {
-            log.warn("Bosh deployment ${id} not found")
+            log.info("[delete]: Bosh deployment ${id} not found")
```
ibonUrrutia: Why are you changing the level from warn to info? In general when you have code released, and people is using it in production, changes like this in the log level can have big impact in production systems. Take care

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -134,25 +139,28 @@ class BoshRestClient {
...
 
-    void postConfig(String config) {
+    BoshConfigResponseDto postConfig(String config) {
         log.debug("Posting new config: \n${config}")
         HttpHeaders headers = createAuthHeaders()
         headers.add('Content-Type', CONTENT_TYPE_JSON)
         HttpEntity<String> request = new HttpEntity<>(config, headers)
-        createRestTemplate().exchange(prependBaseUrl(CONFIGS), HttpMethod.POST, request, String.class)
+
+        return createRestTemplate().exchange(prependBaseUrl(CONFIGS), HttpMethod.POST, request, BoshConfigResponseDto.class)
```
ibonUrrutia: Format this line so max number of character per line are 120

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -134,25 +139,28 @@ class BoshRestClient {
...
+                .body
     }
 
+
```
ibonUrrutia: Format your code before pushing... Usually IDE's are deleting this empty lines which only add noise to the review

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: Agreed, I'm removing `@VisibleForTesting`, keep the method `private` and rather remove the test of this method, as we want to test the API and not the internals.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: Same as before, will remove `@VisibleForTesting`, keep the method `private` and remove the test.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: Makes sense, I'll change the name :+1:

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: Yeah, maybe there will also be more parameters coming.

ibonUrrutia: > Yeah, maybe there will also be more parameters coming.

In that case, rename the `BoshConfigRequestDto` to `BoshConfigurationRequest` and use immutables java library: it won't be just a DTO and using immutables is really easy to extend certain class with more properties and generate all this builder methods.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: Replaced the `Preconditions.checkNotNull` with a `Assert.hasText` which checks for `null` and `empty`. The client is handling when the returned String this method is not okay (`null` or `empty`).

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
ampersand8: As discussed, using the same code style in Intellij and reformatted the code.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshClient.groovy
ampersand8: You're right, reverted back to warn level. Changed it to info by mistake.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: As discussed, using the same code style in Intellij and reformatted the code.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
ampersand8: As discussed, using the same code style in Intellij and reformatted the code.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -15,191 +15,235 @@
...
-    boolean isBoshUndeployTaskSuccessful(LastOperationJobContext lastOperationJobContext) {
-        Optional<String> maybe = ServiceDetailsHelper.from(lastOperationJobContext.serviceInstance.details).findValue(BoshServiceDetailKey.BOSH_TASK_ID_FOR_UNDEPLOY)
-        if (maybe.present) {
-            return isBoshTaskSuccessful(maybe.get())
-        } else {
-            return true
+    private String readTemplateContent(String templateIdentifier, String version = "1.0.0") {
+        Assert.hasText(templateIdentifier, "empty string is invalid template identifier")
+        try {
+            String template = templateConfig.getTemplateForServiceKey(templateIdentifier, version)[0]
```
seilc1: maybe .first() instead of [0] - IMHO a bit more explanatory ... LINQ would have a .single() to make sure that there is only one ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/BoshFacade.groovy
```diff
@@ -15,191 +15,235 @@
...
         }
     }
 
-    void setConfig(String name, String type, String content) {
-        createBoshClient().setConfig(new BoshConfigRequestDto(name: name, type: type, content: content))
-    }
-
-    void deleteConfig(String name, String type) {
-        createBoshClient().deleteConfig(name, type)
+    private static
```
seilc1: linebreak ?!

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -15,164 +15,191 @@
...
-        log.trace("Posting new bosh deployment: \n${data}")
-        HttpHeaders headers = createAuthHeaders()
-        headers.add('Content-Type', CONTENT_TYPE_YAML)
-        HttpEntity<String> request = new HttpEntity<>(data, headers)
-
-        def responseEntity = createRestTemplate().exchange(prependBaseUrl(DEPLOYMENTS), HttpMethod.POST, request, String.class)
-
-        return handleRedirectonAndExtractTaskId(responseEntity)
+    void deleteConfig(String name, String type) {
+        createAuthRestTemplate().exchange(prependBaseUrl(CONFIGS + (new GenericConfigAPIQueryFilter.Builder())
```
seilc1: Shouldn't Builder be a static function? so we do not need the new?

broker/src/test/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProviderSpec.groovy
```diff
@@ -15,42 +15,83 @@
...
+        then:
+        result.isAsync
+        1 * serviceProvider.asyncProvisioningService.scheduleProvision({
+            it.jobClass == ServiceProvisioningJob.class &&
+            it.guid == serviceInstaceGuid && it.retryIntervalInSeconds == serviceProvider.serviceConfig.retryIntervalInSeconds && it.maxRetryDurationInMinutes == serviceProvider.serviceConfig.maxRetryDurationInMinutes
+        })
     }
 
-    def "template customization works correctly"(){
+    def "template customization works correctly"() {
```
seilc1: Could we make a test where we check that the output is really what we expect? compare it with a post replaced templated maybe?

PR #267, opened 2019-06-12
Improve provisioning stability

ampersand8:
We experienced some stability issues during provisioning and deprovisioning. This is sometimes due to failed calls to the shield REST API, for registering and deregistering the service instance for regular backups. Adding retries when calling the API, may help reduce the number of failures during provisioning.

Since Springs `@Retryable` was already used by another component and we therefore already have the dependency, going to also use [Retryable](https://docs.spring.io/spring-retry/docs/api/current/org/springframework/retry/annotation/Retryable.html).

The `@Retryable` feature is tested with WireMock by injecting http failures (status code `500`) to recordings.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -163,19 +230,20 @@ class ShieldClient {
...
+            throw new RuntimeException("Schedule ${shieldServiceConfig.scheduleName.get()} that is configured does not exist on shield")
         }
 
         createOrUpdateJob(jobName, targetUuid, store.uuid, retention.uuid, schedule, paused)
     }
 
     private String createOrUpdateTarget(ShieldTarget target, String targetName, String agent) {
         TargetDto targetOnShield = buildClient().getTargetByName(targetName)
-        targetOnShield == null ? buildClient().createTarget(targetName, target, agent) : buildClient().updateTarget(targetOnShield, target, agent)
+        targetOnShield == null ? buildClient().createTarget(targetName, target, agent) :
```
ibonUrrutia: The day that somebody decides to return an empty `TargetDTO` instead of null from `getTargetByName` we are going to have a hard time trying to guess why we are not able of create or update new targets :-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -186,8 +254,8 @@ class ShieldClient {
                                      boolean paused) {
         JobDto jobOnShield = buildClient().getJobByName(jobName)
         jobOnShield == null ?
```
ibonUrrutia: Same as last comment. The pain of `null`

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientFactory.groovy
```diff
@@ -16,25 +16,36 @@
...
 
+/**
+ * Strategy pattern to support multiple shield API versions
+ * See <a href="https://github.com/shieldproject/shield">Shield Project</a>
+ */
+// TODO: Is not a factory
 @Component
 @CompileStatic
-@Slf4j
 class ShieldRestClientFactory {
```
ibonUrrutia: Man, if it is not a Factory, don't put `TODO`, rename it instead ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientFactory.groovy
```diff
@@ -16,25 +16,36 @@
...
 
     private List<ShieldRestClient> shieldRestClients
 
     @Autowired
     ShieldRestClientFactory(List<ShieldRestClient> shieldRestClients) {
+        Assert.notEmpty(shieldRestClients, "At least one ShieldRestClient has to be provided")
         this.shieldRestClients = shieldRestClients
     }
 
+    /**
```
ibonUrrutia: You are documenting the exact code that I can read two lines below. It would be better describe here WHY we are matching the version, WHY we think that is a hack, and alternatives for the future or plans of refactoring this.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientImpl.groovy
```diff
@@ -21,18 +21,18 @@ import com.swisscom.cloud.sb.broker.backup.shield.dto.*
...
-@Slf4j
 abstract class ShieldRestClientImpl implements ShieldRestClient {
     protected ShieldConfig config
     protected RestTemplate restTemplate
 
-    ShieldRestClientImpl(ShieldConfig shieldConfig, RestTemplateBuilder restTemplateBuilder) {
-        this.restTemplate = restTemplateBuilder.withSSLValidationDisabled().build()
+    ShieldRestClientImpl(ShieldConfig shieldConfig) {
+        Assert.notNull(shieldConfig, "Shield config cannot be null!")
+        this.restTemplate = new RestTemplateBuilder().withSSLValidationDisabled().build()
```
ibonUrrutia: Any reason to introduce this BIG security hole? If it is, document it in the code with a comment.
But against what it looks the consensus in the "it does not matter" developer community, for me, disabling certificate validations in SSL is a BIG problem. So convince me before passing this one.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientv1.groovy
```diff
@@ -17,33 +17,37 @@ package com.swisscom.cloud.sb.broker.backup.shield.restClient
...
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpEntity
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpMethod
 import org.springframework.http.MediaType
 import org.springframework.stereotype.Component
 
-@Slf4j
+/**
+ * {@link ShieldRestClient} implementation for shield version 0.10.x
```
ibonUrrutia: Could you explain WHY are we implementing for such an old version?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientv1.groovy
```diff
@@ -17,33 +17,37 @@ package com.swisscom.cloud.sb.broker.backup.shield.restClient
...
     final int apiVersion = 1
 
     @Autowired
-    ShieldRestClientv1(ShieldConfig shieldConfig, RestTemplateBuilder restTemplateBuilder) {
-        super(shieldConfig, restTemplateBuilder)
+    ShieldRestClientv1(ShieldConfig shieldConfig) {
+        super(shieldConfig)
     }
 
     boolean matchVersion() {
```
ibonUrrutia: Believe or not, this method is using Exceptions for program control flow. Refactor it properly, for example checking if contains a version in response body before comparing it. But I don't want to see any `catch(Exception ex)` in the result... are you sure that the only Exception thrown is the one you are expecting for logging "Not shield API version v1"? What happens when the response.body is null ?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientv2.groovy
```diff
@@ -19,30 +19,34 @@ import com.swisscom.cloud.sb.broker.backup.shield.ShieldConfig
...
-import groovy.util.logging.Slf4j
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.*
 import org.springframework.stereotype.Component
 
-@Slf4j
+/**
+ * {@link ShieldRestClient} implementation for shield version 8.x
```
ibonUrrutia: Again, say WHY we have an implementation for this version.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    def "register and run job"() {
+        given:
+        String jobName = SHIELD_TEST + "-JOB"
+        String targetName = SHIELD_TEST + "-TARGET"
+
+        when:
+        String result = sut.registerAndRunJob(jobName, targetName, shieldTarget, backupParameter, SHIELD_AGENT_URL)
+        backupTaskUUID = result
```
ibonUrrutia: What? Are you overwriting in one test a @Shared member with the output of the method? BAD idea

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
```diff
@@ -0,0 +1,343 @@
...
+
+        then:
+        result.find {it.key == SHIELD_JOB_UUID.key}.value.length() > 0
+        result.find {it.key == SHIELD_TARGET_UUID.key}.value.length() > 0
+    }
+
+    def "unregister system backup"() {
+        given:
+        if (!SHIELD_MOCKED) {
+            sleep(2000)
```
ibonUrrutia: Instead of using flaky sleeps, we should discuss how to test these asynchronous stuff properly with Spock!

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
```diff
@@ -0,0 +1,343 @@
...
+        exception.message == message
+
+        where:
+        taskUUID           | message
+        null               | "Task UUID cannot be empty!"
+        "TASKDOESNOTEXIST" | "Rest call to Shield failed with status:501 NOT_IMPLEMENTED"
+    }
+
+    @Unroll
+    def "register and run system backup throws ServiceBrokerException when given 500 or 504 failure at #failingUrl"() {
```
ibonUrrutia: Test also when the socket is close, when it returns gibberish... use Wiremock failing stuff as I showed in my tests.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
ampersand8: @ibonUrrutia every single test should be independent?

ibonUrrutia: for avoiding false positives and true falses ;-) yes

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/BackupOnShield.groovy
```diff
@@ -71,6 +70,12 @@ trait BackupOnShield extends ExtensionProvider {
         String storageName = planParamtersForBackup.find {
             it.getName().equals("BACKUP_STORAGE_NAME")
         }?.getValue()
-        new BackupParameter(scheduleName: scheduleName, retentionName: policyName, storeName: storageName, schedule: schedule)
+        return backupParameter().
+                scheduleName(Optional.ofNullable(scheduleName)).
```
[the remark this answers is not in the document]
ampersand8: Used `@Value.Default` as suggested in place of `Optional<String>`

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/SystemBackupProvider.groovy
```diff
@@ -39,7 +39,7 @@ trait SystemBackupProvider extends BackupOnShield {
         shieldClient.registerAndRunSystemBackup(jobName, targetName, shieldTarget, shieldServiceConfig, shieldAgentUrl(serviceInstance))
     }
 
-    def unregisterSystemBackupOnShield(String serviceInstanceId) {
+    void unregisterSystemBackupOnShield(String serviceInstanceId) {
```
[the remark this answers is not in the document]
ampersand8: Added logging statement to this method

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/BackupParameter.java
```diff
@@ -13,14 +13,24 @@
...
-    String agent
+import java.util.Optional;
+
+@Value.Immutable
+public interface BackupParameter extends BackupServiceConfig {
+    String getStoreName();
+
+    String getRetentionName();
+
+    Optional<String> getScheduleName();
```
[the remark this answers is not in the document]
ampersand8: Implemented `@Value.Default` as suggested

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/BackupParameter.java
```diff
@@ -13,14 +13,24 @@
...
+
+    Optional<String> getSchedule();
```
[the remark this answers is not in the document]
ampersand8: Addressed as mentioned before

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/BackupParameter.java
```diff
@@ -13,14 +13,24 @@
...
-class BackupParameter implements BackupServiceConfig {
-    String storeName
-    String retentionName
-    String scheduleName
-    String schedule
-    String agent
+import java.util.Optional;
+
+@Value.Immutable
+public interface BackupParameter extends BackupServiceConfig {
```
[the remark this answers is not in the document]
ampersand8: Implemented as abstract class and overwrote the builder to reduce references to `ImmutableBackupParameter` in the code.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientv1.groovy
ampersand8: Changed method to deal with matching the API version more specifically.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/restClient/ShieldRestClientv2.groovy
ampersand8: Added some doc, trying to explain why.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
ampersand8: Implemented tests independent of each other.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
ampersand8: Using a polling mechanism as a simple solution to wait for end states.

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClientTest.groovy
ampersand8: Added more tests regarding the network stability.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -20,66 +20,126 @@ import com.swisscom.cloud.sb.broker.backup.BackupPersistenceService
...
+
+import javax.annotation.PostConstruct
 
 @Component
-@Slf4j
+@CompileStatic
 class ShieldClient {
+    private static final Logger LOG = LoggerFactory.getLogger(ShieldClient.class)
+
     protected ShieldConfig shieldConfig
```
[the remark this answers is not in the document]
ampersand8: Agreed, I'll also set the properties to `private` and the class to `final`, as there's no intention to extend this class.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -20,66 +20,126 @@ import com.swisscom.cloud.sb.broker.backup.BackupPersistenceService
...
         this.shieldConfig = shieldConfig
         this.shieldRestClientFactory = shieldRestClientFactory
-        this.restTemplateBuilder = restTemplateBuilder
+        this.restTemplateBuilder = new RestTemplateBuilder()
         this.backupPersistenceService = backupPersistenceService
     }
 
-    String registerAndRunJob(String jobName, String targetName, ShieldTarget shieldTarget, BackupParameter shieldServiceConfig, String shieldAgentUrl) {
+    @PostConstruct
+    void init() {
```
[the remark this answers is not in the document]
ampersand8: Yeah, makes more sense. KISS to the rescue!

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -20,66 +20,126 @@ import com.swisscom.cloud.sb.broker.backup.BackupPersistenceService
...
+                             BackupParameter backupParameter,
+                             String shieldAgentUrl) {
+        Assert.hasText(jobName, "Job name cannot be empty!")
+        Assert.hasText(targetName, "Target name cannot be empty!")
+        Assert.notNull(shieldTarget, "ShieldTarget cannot be null!")
+        Assert.notNull(backupParameter, "BackupParameter cannot be null!")
+        Assert.hasText(shieldAgentUrl, "Shield agent URL cannot be empty!")
         String targetUuid = createOrUpdateTarget(shieldTarget, targetName, shieldAgentUrl)
-        String jobUuid = registerJob(jobName, targetUuid, shieldServiceConfig)
+        String jobUuid = registerJob(jobName, targetUuid, backupParameter)
```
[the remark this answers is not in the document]
ampersand8: Good idea, added a log statement.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,40 +164,44 @@ class ShieldClient {
...
     void deleteBackup(String taskUuid) {
+        Assert.hasText(taskUuid, "Task UUID cannot be empty!")
         try {
             TaskDto task = buildClient().getTaskByUuid(taskUuid)
             buildClient().deleteArchive(task.archive_uuid)
         }
         catch (ShieldResourceNotFoundException e) {
             // either task or archive is not existing on shield (anymore); probably because it was deleted already
-            log.warn("Could not delete backup because it was not found anymore on Shield; do not fail though", e)
+            LOG.warn("Could not delete backup because it was not found anymore on Shield; do not fail though", e)
```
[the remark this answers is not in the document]
ampersand8: yeah, the method name change makes the intent more clear

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
ampersand8: Fair enough, added a condition to check for the uuid, as any valid target would respond with one set.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
ampersand8: same solution as above

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -20,66 +20,126 @@ import com.swisscom.cloud.sb.broker.backup.BackupPersistenceService
...
+     * we will replace it with a circuit breaker solution.
+     */
+    @Retryable(
+            value = ServiceBrokerException.class,
+            maxAttempts = 3,
+            backoff = @Backoff(delayExpression = "#{\${com.swisscom.cloud.sb.broker.shield.backOffDelay}}"))
+    void unregisterSystemBackup(String jobName, String targetName) {
+        Assert.hasText(jobName, "Job name cannot be empty!")
+        Assert.hasText(targetName, "Target name cannot be empty!")
         deleteJobIfExisting(jobName)
```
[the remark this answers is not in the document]
ampersand8: Would you change the log level to `info`?

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -20,66 +20,126 @@ import com.swisscom.cloud.sb.broker.backup.BackupPersistenceService
...
         deleteTargetIfExisting(targetName)
     }
 
-    def deleteTask(String taskUuid) {
+    void deleteTask(String taskUuid) {
+        Assert.hasText(taskUuid, "Task UUID cannot be empty!")
         buildClient().deleteTaskByUuid(taskUuid)
     }
 
     JobStatus getJobStatus(String taskUuid, Backup backup = null) {
```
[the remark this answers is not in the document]
ampersand8: Yeah, this is something we need to fix. But I'd rather do this outside of this PR as I think there's some logic we need to think about, like retrying a backup job if it fails (which was the goal of this code).

PR #271, opened 2019-07-08
Bug/bosh task ids are not overwritten

seilc1:
define service details as unique, to make sure it gets replaced.


ampersand8:
@seilc1 I added a test for this behaviour, with that this PR is ok with me. But I'm not sure whether we should get entirely rid of the option to not have `uniqueKey` set to false. Is there a use case for that?
And if so, we might want to change the behaviour of [ServiceDetailHelper.getValue()](https://github.com/swisscom/open-service-broker/blob/develop/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/servicedetail/ServiceDetailsHelper.groovy#L58) to e.g. return the latest value by default.

PR #272, opened 2019-07-09
Improve shield API client reliability

ampersand8:
**Problem**
The REST API of the backup system shield is not always responding properly. We observe sometimes responses with status code 500 which can be successfully retried a few seconds later. Also since shield can be in a different network, there might be many reasons a single request fails.

**Proposal**
Every HTTP REST call should be retried if a retryable error happens (this excludes e.g. 404). Currently one of the best solutions comes from resilience4j, which we picked in this PR.

**Note**
Decision was made to no longer support shield version >= 8. Therefore we're dropping the strategy pattern for `ShieldRestClient`.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/SystemBackupProvider.groovy
```diff
@@ -39,7 +39,7 @@ trait SystemBackupProvider extends BackupOnShield {
         shieldClient.registerAndRunSystemBackup(jobName, targetName, shieldTarget, shieldServiceConfig, shieldAgentUrl(serviceInstance))
     }
 
-    def unregisterSystemBackupOnShield(String serviceInstanceId) {
+    void unregisterSystemBackupOnShield(String serviceInstanceId) {
```
ibonUrrutia: in this case, that you had to calculate `targetName`, I would return it in  the method. After unregistering it, is the last opportunity to have it and maybe is useful in some ui for writing successful unregister messages

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldApiException.java
```diff
@@ -0,0 +1,36 @@
...
+    public static ShieldApiException of(String description, Throwable ex) {
+        return new ShieldApiException(description, ex);
+    }
+
+    public boolean isHttpStatusCodeException() {
+        return httpStatus != null;
+    }
+
+    public boolean isNotFound() {
+        return isHttpStatusCodeException() && (httpStatus.value() == 404) || httpStatus.value() == 501;
```
ibonUrrutia:
```suggestion
        return isHttpStatusCodeException() && (httpStatus.value() == 404 || httpStatus.value() == 501);
```
Is what I think what you wanted (but a switch case would be better for adding easily more cases in the future).

What I don't like is that according to this method an HTTP status code of 501 is a _not found_. A _not implemented_ is not the same (if I check the `isNotFound()` and decide to create the resource in that case and receive a _not implemented_ again?).

If it is a specific case of Shield, is a perfect candidate for being explained in the javadoc. So, you decide: document it or change it.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldApiException.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public String toString() {
```
ibonUrrutia: Bad `toString` . IMHO: 
```suggestion
    public String toString() {
               return String.format("ShieldApiException[%s %s: %s]",
                                                  isHttpStatusCodeException() ? httpStatus.value() : "-",
                                                  description != null ? description : "no description",
                                                  getCause() != null ? getCause().getMessage() : "no cause");
```
will be better because is human readable and much easier to be parsed with a script. But what about an immutable value so you don't have to make those checks of nullity? An exception is a perfect candidate for being immutable.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -18,70 +18,102 @@ package com.swisscom.cloud.sb.broker.backup.shield
...
+        this.apiClient = ShieldRestClientV1.of(shieldConfig)
+        this.restTemplateBuilder = new RestTemplateBuilder()
         this.backupPersistenceService = backupPersistenceService
-    }
-
-    String registerAndRunJob(String jobName, String targetName, ShieldTarget shieldTarget, BackupParameter shieldServiceConfig, String shieldAgentUrl) {
+        LOG.info(shieldConfig.toString())
+    }
+
+    String registerAndRunJob(String jobName,
```
ibonUrrutia: This 5 parameters + an UUID make for me an immutable `BackupJob` that you can pass as parameter and return from a method: ```BackupJob run(BackupJob)```

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -18,70 +18,102 @@ package com.swisscom.cloud.sb.broker.backup.shield
...
-        buildClient().runJob(jobUuid)
-    }
-
-    Collection<ServiceDetail> registerAndRunSystemBackup(String jobName, String targetName, ShieldTarget shieldTarget, BackupParameter shieldServiceConfig, String shieldAgentUrl) {
+        String jobUuid = registerJob(jobName, targetUuid, backupParameter)
+        LOG.debug("Running registered backup job '{}' with jobname '{}' on shield", jobUuid, jobName)
+        apiClient.runJob(jobUuid)
+    }
+
+    Collection<ServiceDetail> registerAndRunSystemBackup(String jobName,
```
ibonUrrutia: Again, with `BackupJob` you simplify this interface (you could add a `isSystemJob` for example). And the method would be: `BackupJob runSystem(BackupJob job)`

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -18,70 +18,102 @@ package com.swisscom.cloud.sb.broker.backup.shield
...
+        String jobUuid = registerJob(jobName, targetUuid, backupParameter, false)
+        LOG.debug("Running registered system backup job '{}' with jobname '{}' on shield", jobUuid, jobName)
+        apiClient.runJob(jobUuid)
 
         [ServiceDetail.from(ShieldServiceDetailKey.SHIELD_JOB_UUID, jobUuid),
          ServiceDetail.from(ShieldServiceDetailKey.SHIELD_TARGET_UUID, targetUuid)]
     }
 
-    def unregisterSystemBackup(String jobName, String targetName) {
+    void unregisterSystemBackup(String jobName, String targetName) {
```
ibonUrrutia: Same, `UUID deleteSystem(BackupJob job)` makes sense.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -18,70 +18,102 @@ package com.swisscom.cloud.sb.broker.backup.shield
...
+        Assert.hasText(jobName, "Job name cannot be empty!")
+        Assert.hasText(targetName, "Target name cannot be empty!")
+        LOG.debug("Deregistering Job '{}' and Target '{}' from Shield", jobName, targetName)
         deleteJobIfExisting(jobName)
         deleteTargetIfExisting(targetName)
     }
 
-    def deleteTask(String taskUuid) {
-        buildClient().deleteTaskByUuid(taskUuid)
+    void deleteTask(String taskUuid) {
```
ibonUrrutia: This method tells me that `BackupJob` has the concept of `Task`. Maybe a `BackupJob` is a `Collection` of different `Task` that need to be done

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -18,70 +18,102 @@ package com.swisscom.cloud.sb.broker.backup.shield
...
+        Assert.hasText(taskUuid, "Task UUID cannot be empty!")
+        apiClient.deleteTaskByUuid(taskUuid)
     }
 
+    //FIXME: Do not retry within getJobStatus, as it's not expected
     JobStatus getJobStatus(String taskUuid, Backup backup = null) {
```
ibonUrrutia: And this method tells me that a `BackupJob` has certain `JobStatus`, that I could query with `BackupJob updateStatus(BackupJob job)`. Actually, the implementation tells me that the status of a `BackupJob` is the status of its associated `Tasks`. In fact if I store the different tasks in a `BackupJob` I have a basic log of what happen and what failed.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,78 +136,91 @@ class ShieldClient {
     }
 
     String getJobName(String jobUuid) {
-        buildClient().getJobByUuid(jobUuid).name
+        Assert.hasText(jobUuid, "Job UUID cannot be empty!")
+        apiClient.getJobByUuid(jobUuid).name
     }
 
     String restore(String taskUuid) {
```
ibonUrrutia: Umm, does this method tells me that there is the concept of a `RestoreJob`? ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,78 +136,91 @@ class ShieldClient {
...
+        Assert.hasText(taskUuid, "Task UUID cannot be empty!")
+        TaskDto task = apiClient.getTaskByUuid(taskUuid)
         // TODO need to check for archive, status etc. here?
-        buildClient().restoreArchive(task.archive_uuid)
+        apiClient.restoreArchive(task.archive_uuid)
     }
 
     // avoid throwing exceptions in this method so that users can delete failed backups.
-    void deleteBackup(String taskUuid) {
+    void deleteBackupIfExisting(String taskUuid) {
```
ibonUrrutia: much better to have a `BackupJob` passed and returned, that holds inside the historical of what happened...

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,78 +136,91 @@ class ShieldClient {
...
+            if (e.isNotFound()) {
+                // either task or archive is not existing on shield (anymore); probably because it was deleted already
+                LOG.warn("Could not delete backup because it was not found anymore on Shield; do not fail though", e)
+            } else {
+                throw e
+            }
         }
     }
 
     void deleteJobsAndBackups(String serviceInstanceId) {
```
ibonUrrutia: With `BackupJob` you simplify those three delete methods in one `BackupJob delete(BackupJob job)`

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,78 +136,91 @@ class ShieldClient {
...
-        TargetDto target = buildClient().getTargetByName(targetName)
+    private void deleteTargetIfExisting(String targetName) {
+        TargetDto target = apiClient.getTargetByName(targetName)
         if (target != null) {
-            buildClient().deleteTarget(target.uuid)
+            apiClient.deleteTarget(target.uuid)
         }
     }
 
     private JobStatus statusOfArchive(TaskDto task) {
```
ibonUrrutia: Did I say that you simplify with `BackupJob` also this method ? ;-)

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/backup/shield/ShieldClient.groovy
```diff
@@ -104,78 +136,91 @@ class ShieldClient {
...
+        if (archive != null && archive.status.isValid()) {
             return JobStatus.SUCCESSFUL
         } else {
             return JobStatus.FAILED
         }
     }
 
-    private String registerJob(String jobName, String targetUuid, BackupParameter shieldServiceConfig, boolean paused = true) {
-        StoreDto store = buildClient().getStoreByName(shieldServiceConfig.storeName)
+    private String registerJob(String jobName,
```
ibonUrrutia: and again, again and again :-)

PR #276, opened 2019-07-18
bug: checkAuthTypeAndLogin() returns Optional<String>

ampersand8:
Since `checkAuthTypeAndLogin()` returns `Optional<String>` and not `String`, it should be treated this way, otherwise the token will be invalid, as the `toString()` method will return `Optional[<Actual Token String>]`.
Furthermore the usage of `null` can be avoided, which is good.

This communication would be covered with the `BoshFacadeTest`, unfortunately the error was only detected when bosh was not mocked. By capturing the `Authorization` header as well, the tests now also fail when mocked and a wrong authorization header is given.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/bosh/client/BoshRestClient.groovy
```diff
@@ -101,7 +101,8 @@ class BoshRestClient {
 
     void deleteConfig(String name, String type) {
         createAuthRestTemplate().exchange(prependBaseUrl(CONFIGS + GenericConfigAPIQueryFilter.builder()
-                .withName(name).withType(type).build().asUriString()),
+                                                                                              .withName(name).withType(
```
seilc1: strange lineberak

ampersand8: Yeah, unlucky reformat. Changed it by using static import and shorter names for builder methods.

PR #281, opened 2019-07-28
bug: no failure on bosh error state

ampersand8:
Bosh director REST API returns `state: error` rather than `state: errored` which is not recognized as failure from `BoshFacade` when checking whether task is successful.

Links:
[Bosh Director HTTP API doc](https://bosh.io/docs/director-api-v1/#tasks)


ampersand8:
Will be fixed in `BoshFacade` refactoring

PR #298, opened 2019-09-12
Feature/add multiple identical key handling for inventory

seilc1:
The inventory service didn't handle multiple service details with the same key 'properly'. I rectified this with adding a new function getAll and replaceByKey.

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
@@ -1,6 +1,5 @@
 package com.swisscom.cloud.sb.broker.services.inventory
 
-import com.google.common.base.Preconditions
 import com.swisscom.cloud.sb.broker.model.ServiceDetail
 import com.swisscom.cloud.sb.broker.model.ServiceInstance
 import com.swisscom.cloud.sb.broker.repository.ServiceDetailRepository
@@ -9,9 +8,9 @@ import groovy.transform.CompileStatic
 import org.springframework.data.util.Pair
 import org.springframework.stereotype.Service
 
-import static org.apache.commons.lang.StringUtils.isNotBlank
 import static com.google.common.base.Preconditions.checkArgument
 import static com.google.common.base.Preconditions.checkState
+import static org.apache.commons.lang.StringUtils.isNotBlank
 
 @CompileStatic
 @Service
@@ -23,6 +22,7 @@ class LocalInventoryServiceImpl implements InventoryService {
     public static String ERROR_DETAIL_NOT_FOUND = "no details for key:%s found"
     public static String ERROR_DETAIL_NOT_UNIQUE = "multiple details for key:%s found"
     public static String ERROR_DETAIL_MANDATORY = "details are mandatory"
+    public static String ERROR_DETAIL_BLANK = "detail key may not be blank"
 
     private final ServiceInstanceRepository serviceInstanceRepository
     private final ServiceDetailRepository serviceDetailRepository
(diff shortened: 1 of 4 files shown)


seilc1:
> I'd like to have parameter validation for the public methods, I added a commit with an example test showing the expectations.
> 
> Other than that, the method `get(String serviceInstanceGuid)` might be a little confusing next to `getAll(String serviceInstanceGuid, String key)`. Those could potentially be replaced with `getAllByServiceInstanceGuid(String serviceInstanceGuid)` and `getAllByKey(String serviceInstanceGuid, String key)`. In that case I'd suggest to deprecate the `get(String serviceInstanceGuid)`.

I've added documentation to the interface to clarify the function of the functions ;-) I think the name in combination with the parameters is already saying enough.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
```diff
@@ -25,41 +36,59 @@ class LocalInventoryServiceImpl implements InventoryService {
...
     }
 
     @Override
     Pair<String, String> get(String serviceInstanceGuid, String key, String defaultValue) {
-        def detail = getServiceInstance(serviceInstanceGuid).details.find { d -> d.key.equalsIgnoreCase(key) }
+        checkArgument(isNotBlank(serviceInstanceGuid), ERROR_SERVICE_INSTANCE_ID_NOT_DEFINED)
+        checkArgument(isNotBlank(key), ERROR_KEY_NOT_DEFINED)
+
+        def details = getServiceInstance(serviceInstanceGuid).details.findAll { d -> d.key.equalsIgnoreCase(key) }
+        checkState(details.size() <= 1, ERROR_DETAIL_NOT_UNIQUE, key)
```
ampersand8: Why is it not unique when `size() <= 1`?

broker/core/src/test/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImplSpec.groovy
```diff
@@ -132,7 +128,7 @@ class LocalInventoryServiceImplSpec extends Specification {
         result.second == "default_value_001"
```
ampersand8: I'd for sure also put the behaviour, that the default value is added to the result even when a key is found, to the interface doc.

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
```diff
@@ -25,41 +36,59 @@ class LocalInventoryServiceImpl implements InventoryService {
...
-        def detail = getServiceInstance(serviceInstanceGuid).details.find { d -> d.key.equalsIgnoreCase(key) }
+        checkArgument(isNotBlank(serviceInstanceGuid), ERROR_SERVICE_INSTANCE_ID_NOT_DEFINED)
+        checkArgument(isNotBlank(key), ERROR_KEY_NOT_DEFINED)
 
-        if (detail == null) {
-            ErrorCode.SERVICEBROKERSERVICEPROVIDER_INTERNAL_SERVER_ERROR.throwNew("No serviceDetail with key:${key} found.")
-        }
+        def details = getServiceInstance(serviceInstanceGuid).details.findAll { d -> d.key.equalsIgnoreCase(key) }
+        checkState(details.size() > 0, ERROR_DETAIL_NOT_FOUND, key)
+        checkState(details.size() == 1, ERROR_DETAIL_NOT_UNIQUE, key)
```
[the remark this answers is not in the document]
seilc1: it checks that the size() is == 1 ... so the exception is only thrown if it is != 1 - preconditions are sometimes confusing

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
seilc1: same as above

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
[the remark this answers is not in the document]
ampersand8: Of course, yeah got confused.

PR #320, opened 2019-10-23
Bug/persistence does not work correctly

seilc1:


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ProvisioningPersistenceService.groovy
```diff
@@ -15,14 +15,16 @@
```
ibonUrrutia: Several stuff in this one:
- We are not going to need to maintain ProvisionRequest if using EIP. But is ok for the legacy ones.
- INFO level for each entering call to the method is too much log. I would put it in DEBUG or even TRACE

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/services/inventory/LocalInventoryServiceImpl.groovy
```diff
@@ -16,6 +18,8 @@ import static org.apache.commons.lang.StringUtils.isNotBlank
 @Service
 class LocalInventoryServiceImpl implements InventoryService {
```
ibonUrrutia: Same in this one INFO is too much

PR #321, opened 2019-11-07
fix: Add more useful toString to ServiceDetail

ibonUrrutia:


model/src/main/groovy/com/swisscom/cloud/sb/broker/model/ServiceDetail.groovy
```diff
@@ -48,26 +48,25 @@ class ServiceDetail extends BaseModel{
...
 
 
     @Override
     public String toString() {
-        return new StringJoiner(", ", ServiceDetail.class.getSimpleName() + "[", "]")
-                .add("key='" + key + "'")
-                .add("value='" + value + "'")
-                .add("type='" + type + "'")
-                .add("uniqueKey=" + uniqueKey)
-                .add("serviceInstance=" + serviceInstance)
```
seilc1: why removing the service instance? I say atleast ad the service_instance_id. this helps if we look at an service_instance separate from the service_instance.

ibonUrrutia: yeah, was not sure about it. I'll do

PR #323, opened 2019-11-08
feat: Modularize the metrics in its own module

ibonUrrutia:
We want to be able to deactivate the metrics gathering if there is any
problem with the metrics servers. The easiest way in this moment was
creating a separated module.

WARNING: For modularizing we had to delete any reference to
PlanBasedMetricsService in
com.swisscom.cloud.sb.broker.servicedefinition.ServiceDefinitionProcessor
so that means that metrics gathering is BROKEN with this commit!


seilc1:
I had to define the interfaces in the utils package, as the metrics package already references the database package and having the database package reference metrics would have produced an dependency loop.

Also the NoOpImplementation is only used, when no other Bean is provided. So if you would like to use your own implementation, just provide your own, and the NoOpImplementation is not used anymore - see: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnMissingBean.html

ibonUrrutia:
> 
> Also the NoOpImplementation is only used, when no other Bean is provided. So if you would like to use your own implementation, just provide your own, and the NoOpImplementation is not used anymore 

That is the point: if we are going to use Spring integration metrics we don't need an implementation of this interface. But anyway I'm merging because we have a problem now.

PR #326, opened 2019-12-13
bug: Make CredHubCredential usable

ampersand8:
CredHubCredential is currently not usable, as CredHubService
is directly autowired in CredHubCredential. This does not work
as CredHubService needs to be properly initialized with a
Configuration. This role is now fulfilled with CredentialStoreConfiguration
which takes care of offering a CredentialStore.


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/binding/CredentialStoreConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Configuration
+public class CredentialStoreConfiguration {
```
seilc1: Should we provide an configuration in the framework? Or shouldn't the implementing parties do this? I fear/feel that this might be to restricting.
If we are checking the spring boot repositories the split (auto-)configuration and implementation in two different modules, so that you can but don't have to use their configuration.

ampersand8: Good point, I'd move `CredentialStoreConfiguration` to a new module `configuration` or `autoconfiguration`.
Alternatively it could also be moved to the `broker` module, but then every client would be forced to provide a configuration.
@seilc1 What do you think?

seilc1: i think the configuration module is `good practice` following the spring boot approach. I think one configuration for all is `good enough`

PR #330, opened 2019-12-27
bug: Disallow null values for ServiceContextDetail

ampersand8:
`value` and `key` should not allow any null values
as this leads to problems later on (e.g. running
`hashCode()` throws a `NullPointerException`) and it
makes no sense to have a null value for those fields.


seilc1:
What about adding the `UPDATE service_context_detail SET _value = '' WHERE _value is NULL;` statement as an flyway migration to make sure it's everywhere fixed?

ampersand8:
@seilc1 done, together with the schema change of the table.

PR #331, opened 2019-12-30
fix: switch to optional implementation

seilc1:
Retrieve the service detail as an optional and handle missing values accordingly. in the deprovision case, we will accept it, as we want to roll back and should support deprovisions of incomplete provisions. for provision it should fail with a correct error.

### DIFF
--- a/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
+++ b/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
@@ -268,17 +268,16 @@ class MongoDbEnterpriseServiceProvider
     @Override
     void unbind(UnbindRequest request) {
         try {
-            def groupId = getMongoDbGroupId(request.serviceInstance)
+            String groupId = getMongoDbGroupId(request.serviceInstance)
+                    .orElseThrow({ ErrorCode.SERVICEPROVIDER_INTERNAL_ERROR.throwNew("MongoDbGroupId is missing, contact support") })
 
-            if (groupId.isPresent()) {
-                opsManagerFacade.checkAndRetryForOnGoingChanges(groupId.get())
-                opsManagerFacade.deleteDbUser(groupId.get(),
-                        ServiceDetailsHelper.from(request.binding.details).getValue(ServiceDetailKey.USER),
-                        ServiceDetailsHelper.from(request.serviceInstance.details).getValue(ServiceDetailKey.DATABASE))
-                opsManagerFacade.deleteOpsManagerUser(ServiceDetailsHelper.from(request.binding.details).
-                        getValue(MongoDbEnterpriseServiceDetailKey.MONGODB_ENTERPRISE_OPS_MANAGER_USER_ID))
-                opsManagerFacade.checkAndRetryForOnGoingChanges(groupId.get())
-            }
+            opsManagerFacade.checkAndRetryForOnGoingChanges(groupId)
+            opsManagerFacade.deleteDbUser(groupId,
+                    ServiceDetailsHelper.from(request.binding.details).getValue(ServiceDetailKey.USER),
+                    ServiceDetailsHelper.from(request.serviceInstance.details).getValue(ServiceDetailKey.DATABASE))
+            opsManagerFacade.deleteOpsManagerUser(ServiceDetailsHelper.from(request.binding.details).
+                    getValue(MongoDbEnterpriseServiceDetailKey.MONGODB_ENTERPRISE_OPS_MANAGER_USER_ID))
+            opsManagerFacade.checkAndRetryForOnGoingChanges(groupId)
         } catch (HttpClientErrorException e) {
             if (e.statusCode == HttpStatus.NOT_FOUND) {
                 log.info(this.getClass().getSimpleName() + ".unbind(): Ignore 404 error during unbind")
--- a/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
+++ b/broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
@@ -36,6 +36,7 @@ import com.swisscom.cloud.sb.broker.services.bosh.BoshTemplate
 import com.swisscom.cloud.sb.broker.services.bosh.BoshTemplateCustomizer
 import com.swisscom.cloud.sb.broker.services.bosh.statemachine.BoshStateMachineFactory
 import com.swisscom.cloud.sb.broker.services.mongodb.enterprise.opsmanager.DbUserCredentials
+import com.swisscom.cloud.sb.broker.services.mongodb.enterprise.opsmanager.OpsManagerCredentials
 import com.swisscom.cloud.sb.broker.services.mongodb.enterprise.opsmanager.OpsManagerFacade
 import com.swisscom.cloud.sb.broker.services.mongodb.enterprise.statemachine.MongoDbEnterperiseStateMachineContext
 import com.swisscom.cloud.sb.broker.services.mongodb.enterprise.statemachine.MongoDbEnterpriseDeprovisionState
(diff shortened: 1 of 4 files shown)


broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/statemachine/MongoDbEnterpriseProvisionState.groovy
```diff
@@ -184,8 +189,8 @@ enum MongoDbEnterpriseProvisionState implements ServiceStateWithAction<MongoDbEn
     DELETE_DEFAULT_ALERTS(LastOperation.Status.IN_PROGRESS, new OnStateChange<MongoDbEnterperiseStateMachineContext>() {
         @Override
         StateChangeActionResult triggerAction(MongoDbEnterperiseStateMachineContext stateContext) {
-            String groupId = MongoDbEnterpriseServiceProvider.getMongoDbGroupId(stateContext.lastOperationJobContext)
-            return new StateChangeActionResult(go2NextState: stateContext.opsManagerFacade.deleteDefaultAlerts(groupId))
+            def groupId = MongoDbEnterpriseServiceProvider.getMongoDbGroupId(stateContext.lastOperationJobContext)
```
ampersand8: Since we declared everywhere else `String`, I'd probably set here `Optional<String>` rather than `def` to make it easy readable. If you really really prefer `def` then I'd at least change the name of the variable as above to `optionalGroupId`, but personally I prefer having the type which then is also the truth.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
```diff
@@ -268,15 +269,16 @@ class MongoDbEnterpriseServiceProvider
...
-            opsManagerFacade.deleteDbUser(groupId,
-                                          ServiceDetailsHelper.from(request.binding.details).
-                                                  getValue(ServiceDetailKey.USER),
-                                          ServiceDetailsHelper.from(request.serviceInstance.details).
-                                                  getValue(ServiceDetailKey.DATABASE))
-            opsManagerFacade.deleteOpsManagerUser(ServiceDetailsHelper.from(request.binding.details).
-                    getValue(MongoDbEnterpriseServiceDetailKey.MONGODB_ENTERPRISE_OPS_MANAGER_USER_ID))
-            opsManagerFacade.checkAndRetryForOnGoingChanges(groupId)
+
+            if (groupId.isPresent()) {
```
ampersand8: I think it's a good idea to allow unbinding when mongodb group id is missing, as something is wrong anyways and binding should not have been possible. But that also implies something went really wrong before, and I'd therefore add an error log in the else statement.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/statemachine/MongoDbEnterpriseDeprovisionState.groovy
```diff
@@ -34,25 +34,32 @@ enum MongoDbEnterpriseDeprovisionState implements ServiceStateWithAction<MongoDb
     DISABLE_BACKUP_IF_ENABLED(LastOperation.Status.IN_PROGRESS, new OnStateChange<MongoDbEnterperiseStateMachineContext>() {
         @Override
         StateChangeActionResult triggerAction(MongoDbEnterperiseStateMachineContext context) {
+            def optionalGroupId = MongoDbEnterpriseServiceProvider.getMongoDbGroupId(context.lastOperationJobContext)
```
ampersand8: Personally I'd avoid using `def` and would prefer `Optional<String>` so that anyone reading the code, knows what type this variable is. The prefix `optional` in the variable name helps, but is not the truth as the type might change in the future and optional might be semantically misunderstood, as in "it's not mandatory to have the group id".

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
```diff
@@ -286,12 +288,12 @@ class MongoDbEnterpriseServiceProvider
         }
     }
 
-    static String getMongoDbGroupId(LastOperationJobContext context) {
+    static java.util.Optional<String> getMongoDbGroupId(LastOperationJobContext context) {
         return getMongoDbGroupId(context.serviceInstance)
     }
 
-    static String getMongoDbGroupId(ServiceInstance serviceInstance) {
+    static java.util.Optional<String> getMongoDbGroupId(ServiceInstance serviceInstance) {
```
ampersand8: Personally I take a little bit issue with those two `getMongoDbGroupId()` methods. The only thing they actually need is `Collection<ServiceDetail>`  but for some reason we give them a whole `LastOperationJobContext` and a `ServiceInstance`. If you wanna boy scout, I'd suggest to replace this method with one just taking the `Collection<ServiceDetail>`, but not necessary.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/statemachine/MongoDbEnterpriseProvisionState.groovy
seilc1: fixed

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
seilc1: will throw if it's missing

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/statemachine/MongoDbEnterpriseDeprovisionState.groovy
seilc1: fixed

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
```diff
@@ -268,15 +269,16 @@ class MongoDbEnterpriseServiceProvider
...
+
+            if (groupId.isPresent()) {
+                opsManagerFacade.checkAndRetryForOnGoingChanges(groupId.get())
+                opsManagerFacade.deleteDbUser(groupId.get(),
+                        ServiceDetailsHelper.from(request.binding.details).getValue(ServiceDetailKey.USER),
+                        ServiceDetailsHelper.from(request.serviceInstance.details).getValue(ServiceDetailKey.DATABASE))
+                opsManagerFacade.deleteOpsManagerUser(ServiceDetailsHelper.from(request.binding.details).
+                        getValue(MongoDbEnterpriseServiceDetailKey.MONGODB_ENTERPRISE_OPS_MANAGER_USER_ID))
+                opsManagerFacade.checkAndRetryForOnGoingChanges(groupId.get())
+            }
```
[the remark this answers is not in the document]
seilc1: see above - it's now less friendly and will throw instead.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
seilc1: I agree from the perspective of making the functions as simple as possible. But to be honest, I see to big a risk in the refactoring of this function due to the whole project not being type safe and missing unit tests.

seilc1: I marked the old functions as deprecated and created a new function with minimal concern.

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
```diff
@@ -268,7 +268,7 @@ class MongoDbEnterpriseServiceProvider
     @Override
     void unbind(UnbindRequest request) {
         try {
-            String groupId = getMongoDbGroupId(request.serviceInstance)
+            String groupId = getMongoDbGroupId(request.serviceInstance.details)
```
Lafunamor: is this automatically returning the value instead of the Optional Object?

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
```diff
@@ -238,7 +238,7 @@ class MongoDbEnterpriseServiceProvider
         def database = ServiceDetailsHelper.from(request.serviceInstance.details).getValue(ServiceDetailKey.DATABASE)
         def hosts = ServiceDetailsHelper.from(request.serviceInstance.details).
                 findAllWithServiceDetailType(ServiceDetailType.HOST)
-        def groupId = getMongoDbGroupId(request.serviceInstance)
+        def groupId = getMongoDbGroupId(request.serviceInstance.details)
```
Lafunamor: please add explicit definition

broker/src/main/groovy/com/swisscom/cloud/sb/broker/services/mongodb/enterprise/MongoDbEnterpriseServiceProvider.groovy
seilc1: on the following Line is an orElseThrow() which handles the null possibility

PR #334, opened 2020-01-23
feat: Allow purging a service instance

ampersand8:
Sometimes there's an unforeseen problem in the
backend which is hard to restore and not necessary
due to the stage used (e.g. testing). For those cases
the operator should have an easy way to remove the
service instance in the service broker database.

With this change a service instance can be easily purged by a REST call to `DELETE /admin/service_instances/{serviceInstanceGuid}/purge`

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
@@ -15,50 +15,111 @@
 
 package com.swisscom.cloud.sb.broker.provisioning
 
+import com.google.common.base.Preconditions
 import com.swisscom.cloud.sb.broker.model.LastOperation
+import com.swisscom.cloud.sb.broker.model.ServiceBinding
 import com.swisscom.cloud.sb.broker.model.ServiceInstance
-import com.swisscom.cloud.sb.broker.repository.ServiceInstanceRepository
 import com.swisscom.cloud.sb.broker.provisioning.lastoperation.LastOperationPersistenceService
+import com.swisscom.cloud.sb.broker.repository.LastOperationRepository
+import com.swisscom.cloud.sb.broker.repository.ServiceInstanceRepository
+import com.swisscom.cloud.sb.broker.services.credential.CredentialStore
 import com.swisscom.cloud.sb.broker.util.Audit
 import groovy.transform.CompileStatic
-import groovy.util.logging.Slf4j
+import org.apache.commons.lang3.StringUtils
 import org.joda.time.LocalDateTime
-import org.springframework.beans.factory.annotation.Autowired
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
 
+import static com.google.common.base.Preconditions.checkArgument
+
 @Component
 @CompileStatic
-@Slf4j
 @Transactional
 class ServiceInstanceCleanup {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ServiceInstanceCleanup.class)
     public static final int MONTHS_TO_KEEP_DELETED_INSTANCE_REFERENCES = 3
 
-    @Autowired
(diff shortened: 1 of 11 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
```diff
@@ -15,50 +15,111 @@
...
-    @Autowired
-    private ProvisioningPersistenceService provisioningPersistenceService
-
-    @Autowired
-    private ServiceInstanceRepository serviceInstanceRepository
+    private final ProvisioningPersistenceService provisioningPersistenceService
+    private final ServiceInstanceRepository serviceInstanceRepository
+    private final LastOperationPersistenceService lastOperationPersistenceService
+    private final LastOperationRepository lastOperationRepository
+    private final CredentialStore credentialStore
```
seilc1: this is never inited!

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
```diff
@@ -15,50 +15,111 @@
...
+        try {
+            serviceInstanceToPurge.
+                    getBindings().
+                    forEach({binding -> deleteCredentialInCredHub(binding as ServiceBinding)})
+        } catch (Exception e) {
+            LOGGER.error("Ignoring any CredHub problems while purging a service instance. Got following exception:", e)
+        }
+
+        provisioningPersistenceService.markServiceInstanceAsDeleted(serviceInstanceToPurge)
+        setSuccessfulDeprovisionLastOperation(serviceInstanceGuid)
```
seilc1: we should make sure that it is traceable that this instance was deleted via an admin action. I suggest doing some [Audit Logging](https://github.com/swisscom/open-service-broker/blob/develop/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/util/Audit.groovy) and maybe store in the message of the LastOperation that this instance was deleted via admin action?

ampersand8: agreed, will add

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
ampersand8: fixed

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
ampersand8: done

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
```diff
@@ -15,50 +15,115 @@
...
+        try {
+            serviceInstanceToPurge.
+                    getBindings().
+                    forEach({binding ->
+                        serviceBindingPersistenceService.delete(binding as ServiceBinding,
+                                                                serviceInstanceToPurge)
+                    })
+        } catch (Exception e) {
+            LOGGER.error("Ignoring any unbinding problems while purging a service instance. Got following exception:", e)
```
seilc1: Suppose this happens. the "Purge" is green ... but this service instance will never be deleted, as it has an undeleted Binding? - IMHO we shouldn't catch here ... this is something which we should check immediately, because else we have dead corpses in the DB.

ampersand8: This is to make sure the purge is not hung up if there's anything going on with the CredentialStore (not found is already handled). The binding in the db would again be tried to be removed when the cleanup job runs through. I'm not sure how permissive this should be.

PR #335, opened 2020-01-27
feat: Deregister Backup

ampersand8:
When a service instance is being purged
an attempt should be made to deregister
the service instance from the backup
system.

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
@@ -88,7 +88,8 @@ class ServiceInstanceCleanup {
     }
 
     /**
-     * Marks a service instance for cleanup and removes any CredHub credentials from bindings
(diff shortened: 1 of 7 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
```diff
@@ -65,7 +72,7 @@ class ServiceInstanceCleanup {
                                                      LastOperation.Status.SUCCESS,
                                                      deleteOlderThan)
         def candidateCount = oprhanedServiceInstances.size()
-        LOGGER.info("Found ${candidateCount} serviceInstance candidate(s) to clean up!")
+        LOGGER.info("Found " + candidateCount + " serviceInstance candidate(s) to clean up!")
```
seilc1: why not proper templating? `"Found {} serviceInstance candidate(s) to clean up!", candidateCount`

ampersand8: done

PR #337, opened 2020-01-27
bug: Load Service eagerly for Plan

ampersand8:
Service is under certain circumstances not
loaded due to the Lazy loading.

Improve testing time for integrationTests

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
@@ -132,16 +132,16 @@ class ServiceInstanceCleanup {
                 errors.add("Failed to delete binding " + binding.getGuid())
             }
         }
-        Pair<Boolean, String> deregisterdFromBackup = deregisterFromBackup(serviceInstanceToPurge.getPlan(),
-                                                                           serviceInstanceGuid)
-        if (!deregisterdFromBackup.getRight().isEmpty()) {
-            errors.add(deregisterdFromBackup.getRight())
+        Pair<Boolean, String> deregisteredFromBackup = deregisterFromBackup(serviceInstanceToPurge.getPlan(),
+                                                                            serviceInstanceGuid)
+        if (!deregisteredFromBackup.getRight().isEmpty()) {
+            errors.add(deregisteredFromBackup.getRight())
         }
 
         return result.
                 purgedServiceInstanceGuid(serviceInstanceGuid).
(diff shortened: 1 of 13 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
```diff
@@ -160,11 +164,13 @@ class ServiceInstanceCleanup {
...
      * @param plan
      * @param serviceInstanceGuid
-     * @return ([Boolean specifying whether ServiceProvider is SystemBackupProvider],
-     * [String containing errors, empty String if no errors occurred])
+     * @return ([BooleanspecifyingwhetherServiceProviderisSystemBackupProvider],
+     *[Stringcontainingerrors,emptyStringifnoerrorsoccurred])
      */
     private Pair<Boolean, String> deregisterFromBackup(Plan plan, String serviceInstanceGuid) {
-        ServiceProvider serviceProvider = serviceProviderLookup.findServiceProvider(plan)
+        Plan planWithService = planRepository.findByGuidAndFetchServiceEagerly(plan.getGuid())
```
seilc1: Should we do a null check here? or maybe return an Optional<Plan> so that we can do the "orElseThrow()"?

broker/src/integration-test/groovy/com/swisscom/cloud/sb/broker/backup/SystemBackupProviderTest.groovy
```diff
@@ -26,21 +27,14 @@ import org.junit.ClassRule
...
 import static com.github.tomakehurst.wiremock.client.WireMock.*
 import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options
 import static org.junit.Assert.assertEquals
 
-@SpringBootTest
-@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
-@ActiveProfiles("default,test,secrets")
 @Stepwise
-class SystemBackupProviderTest extends Specification {
+class SystemBackupProviderTest extends BaseSpecification {
```
seilc1: Thanks for the change to BaseSpecification that removes a lot of duplicated code *thumps up*

broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/provisioning/ServiceInstanceCleanup.groovy
ampersand8: Using a return with an error explaining what went wrong, as deregistering from the backup system is seen as optional and lead to a complete purging failure.

PR #341, opened 2020-01-30
feat: Add Debug Logs for Rest Template

Lafunamor:
This code add additional debug logging to all resttemplate calls.
This enables us to log request and response headers and bodies for
debugging purposes.


seilc1:
will need to check the functional tests.

PR #348, opened 2020-03-09
bug: Log Sensitive Parameters As Confidential

ampersand8:
Logging sensitive parameters as null can be misleading
as you cannot know whether any or which parameters are set.
With this change all top-level parameter keys are shown.

### DIFF
--- a/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/controller/BaseController.groovy
+++ b/broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/controller/BaseController.groovy
@@ -77,4 +77,9 @@ abstract class BaseController {
             it.getDefaultMessage()
         }).join(", ")
     }
+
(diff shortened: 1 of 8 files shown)


broker/core/src/main/groovy/com/swisscom/cloud/sb/broker/controller/BaseController.groovy
```diff
@@ -77,4 +77,9 @@ abstract class BaseController {
             it.getDefaultMessage()
         }).join(", ")
     }
+
+    public static String getSensitiveParameters(Map<String, Object> parameters) {
```
seilc1: wondering if it should be in the `BaseController` or `ControllerHelper` - Ideal would be IMHO in `SensitivieParameterProvider` but Groovy does not support static/default methods in interfaces. I don't think its wrong in the `BaseController` - Therfor approved.

PR #369, opened 2020-10-20
bug: RestTemplateBuilder With Auth Thread Unsafe

ampersand8:
Using RestTemplateBuilder `withBearerAuthentication` or
`withBasicAuthentication`  can lead to `ConcurrentModificationException`
due to the modification of the interceptors. Using
`@Synchronized` resolves this issue.


furban51 (approved):
approved!

